PR #6, opened 2024-04-29
Bitbucket Wiki -> Read the Docs 

robelgeda:
**Warning: this PR is opened against the `stable` branch, not the `dev` branch at the moment.**

#### [RTD DEMO [Private]](https://ramses-robel.readthedocs.io/en/docs/)

This was developed during RUM24 hackathons regarding this issue. I think its a good start and can be merged if we think this is the right direction. 

- I was able to port the ramses BitBucket wiki and add it to the `doc` folder in the code. This was possible because the wiki files are already markdown files. Now we can edit and keep track of the code at the same time. 
    - Sphinx uses RST format so I had to use [`myst-parser`](https://myst-parser.readthedocs.io/en/latest/) to enable markdowns. This is now in the config for handling of future markdown files.
- I removed the auto testing stuff because it was taking up space and we will be using GitHub in the future to do the same task.
- I added a demo on RTD at the link above. This is in my private rtd and I have set it to hidden so only people with links can see it. 
- **We are no longer going to support the LaTex version of the docs, we are going to use the version in this PR, i.e the `.md` files to generate a pdf version of the documentation.** See discussion below.
- If this looks good, upon merging this PR, the only thing that is needed is:
    - One of the admins to create a read the docs account using your GitHub account as login, i.e create the account using your GitHub account. 
    -  Click on the `Import Project` button and select this repo. 
    - Thats it! (Also remember to add multiple people as admins after setup)
    - Other considerations is you can choose which branches to make visible. We should start with only the `stable` assuming this PR goes into `stable`.

## Documentation Summary

Some notes about the Sphinx and Read-The-Docs (RTD) documentation builders:

- The documentation has two sections, the user documentation section with files located in the 'wiki' directory and developer documentation in the `dev_docs` directory. When a PDF is generated, only the wiki version is included. 
- Requirements for the docs can be installed via `pip install -r doc/rtd_requirements.txt`. 
    - For MacOS pdf builds you may need to install `pango` as follows `brew install pango`. See `sphinx_simplepdf` docs at the link below. 
- The build can be configured in `doc/conf.py`.
     - To build pdf locally run `sphinx-build -M html doc` or in the `doc` folder `make html`. The html will be built in `_build/html/index.html`
- To use markdown files, i.e `.md` files, the `myst_parser` extension is used. [Documentation for  `myst_parser`](https://myst-parser.readthedocs.io/en/latest/).
- To convert the docs to PDF, `sphinx_simplepdf` extension is used.
    - The RTD build is configured in `.readthedocs.yaml` under the `commands` list. 
    - If `sphinx_simplepdf` ever breaks, you can remove the custom build instructions there. 
    - To build pdf locally run `sphinx-build -M simplepdf doc`.
    - [Documentation for  `sphinx_simplepdf`](https://sphinx-simplepdf.readthedocs.io/en/latest/index.html).
- As of this version, the color scheme for the docs are the following (can be changed in `doc/conf.py`):
    - `primary_color = '#3061f3'`
    - `secondary_color = '#1A43A4'`
    - `text_color = '#ffffff'`


robelgeda:
On second thought, if this sounds okay, I will just clean up the space in the wiki markdown file names (it will only take a few minutes) before it is merged. For now I have disabled links to files with spaces in them.

doc/wiki/ramses_ug.pdf
rteyssier, line 1: Shall we keep the binary PDF of the user's guide?

mtrebitsch: I'd vote yes for now : I've always found it super useful to have a "static" doc available offline.

Later down the line, when we are confident the new documentation is what we want, we could always generate a PDF from the markdown files (e.g. with pandoc, maybe as one of the github actions) and ship it with the code.

cphyc: I would vote *against* having the binary PDF, because it would clutter the git history. However, we can attach the manual to each release of the code (see e.g. https://github.com/pynbody/genetIC/releases/tag/v1.4.0).

cphyc (commented):
Thanks for the work, this is great!

Is having the whole wiki in the main RAMSES repository a good idea? What do think @rteyssier? Another option would be to have a `ramses-doc` repository containing only the current wiki.

The motivation is to keep the RAMSES repository as lightweight as possible; if we start including the wiki, we'll end up with svgs, pngs, etc. that use a lot of space and make cloning slower.

In addition, how much work would it be @robelgeda to have each `.md` file appear in the toc on the left? At the moment, if I want to access this page (https://ramses-robel.readthedocs.io/en/docs/wiki/Tracers.html), I have to first click on "Chapter 2" but it could very well appear in the toc, couldn't it?

rteyssier:
Yes I agree. Let’s see what other people think but I would favor another repo inside the ramses-organisation.
Having a clear link to the doc repo inside the ramses repo would help people find it and clone it at will.
I think also we should remove the pdf from the repo and just explain that the Makefile should be compiled to generate it.

rteyssier:
After reading the various comments, I have the feeling we should have:
1- a separate repo called "user-guide" or "manual" within the ramses-organisation that contains all the files and the makefile for the user's guide that are used to generate the PDF (but without the PDF itself).
2- a lightweight doc in the ramses repo that contains only markdown files with the basic and also probably most up-to-date documentation. This one could also be based in sphinx-fortran or something similar.
What do you think?

robelgeda:
I think this sounds good, and in practice we would be moving the src folder into the new documentation repo. The src folder contains LaxTex files with a make file. 

That being said, I also question if the pdf version is really needed if we put all the documentation into the .md files. We could separate the two (pdf version and wiki version) as separate sections in the docs. The doc can be built locally using the following command without the need for read-the-docs:
```
pip install -r  rtd_requirements.txt
make html
open _build/html/index.html
```
Then the user can navigate the docs locally. I think having all the docs in one place would be a good idea unless the PDF is a must. Also it is possible to convert the docs in this PR into pdfs automatically if we want to provide a PDF version. If this sounds good, the next step would be to move any info in the pdf + images that is not in the wiki so we can remove the src folder.

robelgeda:
Sorry about the delay, just got a chance to clean up some of the file names and update other quick things that will go a long way. The only thing left to look at the .md files because some of them are not used anywhere and it might be better to get rid of them before we squash and merge anything (we dont want them in the git history if they are never used). You can check out the final draft using the same link above! I think its polished and I have now added all the changes I wanted to.
@rteyssier @cphyc

blaizot:
Hi all, 
Thanks @robelgeda for the great work !!
I like very much the idea of a separate projet "user-guide" within the ramses-organisation. 
I think we should have a unique documentation, so i'd vote for @robelgeda's last suggestion (as i understand it) : remove the doc from the ramses repo and have a single version of that in the user-guide (i.e. no .tex files that will be out of sync with the .md files before we know it).
The local generation of the html documentation that can be read offline is great. People may still want a pdf with the full documentation, though. We should either provide this (for tagged versions) or add a possibility to generate the pdf from the .md files. Is that something possible ? (I see there is a possibility to download a pdf on the current read-the-docs page, but that only gives the current page...)

robelgeda:
@blaizot: Ah, you understood me perfectly and this makes sense, I didn't realize that people would want the PDF. In this case it is possible to do it using only the md files! Here is a demo (it came out way nicer than I thought):
~~[Ramses.pdf](https://github.com/ramses-organisation/ramses/files/15187328/Ramses.pdf)~~ (see update below)

It can be auto generated every build of the docs (i.e every time a PR gets merged) or by hand at every release. Where to put the PDF after its generated would be a good question to ask ourselves.

robelgeda:
After some wrestling I got it to build automatically on RTD!!! I added a link in the front page, at the end of the intro paragraph where it says  "A pdf version can be found [here](https://ramses-robel.readthedocs.io/en/docs/Ramses.pdf)." Its so fast that I don't think we should build by hand. This is nice because we don't need to save a copy of the PDF anywhere but RTD, where it will be built, so no worries about file sizes. Also every version of the pdf will be available with each git-tag on RTD.

rteyssier:
This sounds like the solution. @robelgeda can you summarize what you have in your pull request after all these discussions? I understand that we have now a local doc in the ramses repo with only markdown files and an easy way to generate a PDF outside of the repo if required. Am I right? Shall I now merge?

robelgeda:
Yes! Once the tests are done, we can merge!  I cleaned up my commits and updated the intro above. @rteyssier you will need to create an account on RTD and add this repo once this is merged to activate doc building. Also, please add one or two additional admins on RTD to be safe. Its good for PyPi and RTD to have multiple owners since they are strict due to security.

rteyssier (approved):

cphyc:
@meeseeksdev backport to dev

PR #7, opened 2024-04-30
Run automatically test suite

cphyc:
This ports back the automatic test suite to run on any PR open and merge.

On a side note, the tests take about ~30min to run on two cores, and the repo has 2,000 free minutes / month (see https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes), so we're good to run them approximately ~60× per month.


cphyc:
@rteyssier I've included the test that was running daily as a requirement for the acceptance of any future PR.

Regarding the failing test, it looks like the `barotrop` test fails on 6060289 with an error of 3.4e-12. Should we update the results?

The PDF and the log can be accessed as follows (for this PR and all future ones).
image
image
image

rteyssier:
Do we store the pdf and the log? This was a problem on the bitbucket repo and causing the wiki repo to inflate rapidly.

rteyssier:
Yes I think we can update the reference data for the barotrop test.
Ideally, we want to use the data generated on the GitHub architecture using 2 cores.

cphyc:
Both are getting destroyed after 90 days by default, so it should be fine. Unlike bitbucket, we can remove those (called artifacts on GitHub). However, this means we do not keep the logs for tests older than 90days!

cphyc:
> Yes I think we can update the reference data for the barotrop test. Ideally, we want to use the data generated on the GitHub architecture using 2 cores.

OK, I'm not sure I'll have time to update the answers in the coming weeks, so would you be happy if I just adapted the barotrop one and defer the point above to the future?

rteyssier:
> > Yes I think we can update the reference data for the barotrop test. Ideally, we want to use the data generated on the GitHub architecture using 2 cores.
> 
> OK, I'm not sure I'll have time to update the answers in the coming weeks, so would you be happy if I just adapted the barotrop one and defer the point above to the future?

Yes of course. Go ahead and let me know when I can merge.

cphyc:
Assuming the tests pass, this should be good to go.

rteyssier (approved):

PR #8, opened 2024-05-02
Doc Patch

robelgeda:
Patch for #6 
- Quick patch, I forgot to update the path in the front page README, its currently pointing to where the LaTex folder was.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@
 [8]: https://github.com/pynbody/pynbody
 [9]: https://yt-project.org
 
-![GitHub logo dark-mode-only](./doc/src/img/full_logo_white.svg#gh-dark-mode-only)
-![GitHub logo light-mode-only](./doc/src/img/full_logo.svg#gh-light-mode-only)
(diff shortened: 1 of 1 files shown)


mtrebitsch (approved):
Looks good to me (both files exist)!

rteyssier (approved):

blaizot:
Hey ! 
This all looks great. Congrats !! 
Just two minor points: 
- should we not name the "wiki" directory "for-users" (or user_doc, or anything explicit like this) ?
- the file docs.md in doc/dev_docs could be renamed README.md and placed in doc/. directly so that it is displayed on the web interface when we open the doc directory. That would be very neat. (That would also leave the directory dev_docs free for a future documentation for developers?).

robelgeda:
@blaizot Thank you! 
I like the idea of having a README and will add it in the next patch. I think renaming "wiki" will require some carful refactoring since things link there (it might not be as bad, but I not sure).

cphyc:
@meeseeksdev backport to dev

cphyc:
@meeseeksdev backport to dev

PR #12, opened 2024-05-08
Remove whitespaces

cphyc:
This does two things:
- Add a "pre-commit" hook that automatically removes trailing whitespaces,
- Apply it to all files in the trunk.

The pre-commit hook can be installed on the developer side and/or run on GitHub. These are essentially small scripts being run whenever you do "git commit"  to enforce some basic formatting, here removing trailing whitespaces. They can also run automatically against any pull request (we'd have to activate the free service though) so that developers do not have anything to do.

Regarding this PR, and assuming you're happy with this change, we can either add the hook and *not* apply it (so the cleaning up will be done gradually as time proceeds, so essentially only accept f949c73) or apply it once and for all (this PR).


rteyssier:
@meeseeksdev backport to stable

cphyc:
@meeseeksdev backport to stable

PR #13, opened 2024-05-08
Autoformat with Fortran-linter

cphyc:
This applies an opinionated [Fortran linter](https://github.com/cphyc/fortran-linter) on the code base. It should not affect any of the code but makes the formatting uniform.

Note that, as in #12, this could be added as a pre-commit and run only on the files that are part of a changeset for each PR, so that the modifications are gradually applied rather than this massive and impossible-to-review PR.


rteyssier:
Any news on this @cphyc ?

cphyc:
Sure.

After having applied the autoformatter, the code still runs fine (all tests pass, see #16). 
I can propose the following:
- if you think this would be useful, then proceed to next step
- have a look at some of the files from this PR and tell me
  - whether you see anything fishy,
  - whether you think the formatting makes the code better.

I recommend hiding the whitespace if you're looking at the code on GitHub to make it less cluttered.

Before we merge though, I will need to implement an option so that the autoformatter doesn't complain when the file length is too long! Currently, the autoformatter is idempotent but still complains about lines that are too long (because it cannot *yet* split them).

rteyssier:
Literary all files have been changed massively.
It will be very hard to detect a bug in this pull request.

Question: is this formatting consistent with emacs fortran-mode?

Romain

nbrucy:
You may need to rebase the PR onto dev :)

PR #36, opened 2024-10-28
Global parameters for the different number of hydro/mhd variables

tinecolman:
Introduce `nvar_all`, `neul` and `nhydro` variables to simplify differences between MHD versus HYDRO. This allows for more general code and less preprocessor statements.

### DIFF
--- a/hydro/hydro_parameters.f90
+++ b/hydro/hydro_parameters.f90
@@ -49,8 +49,13 @@ module hydro_parameters
   real(dp),dimension(1:MAXBOUND)::u_bound=0
   real(dp),dimension(1:MAXBOUND)::v_bound=0
   real(dp),dimension(1:MAXBOUND)::w_bound=0
-  ! TODO allow for other variables in inflow?
-
+  ! TODO allow other variables in inflow:
+#if NENER>0
+  real(dp),dimension(1:MAXBOUND,1:NENER)::prad_bound=0
+#endif
+#if NVAR>NHYDRO+NENER
+  real(dp),dimension(1:MAXBOUND,1:NVAR-NHYDRO-NENER)::var_bound=0
+#endif
   ! Refinement parameters for hydro
   real(dp)::err_grad_d=-1.0d0  ! Density gradient
   real(dp)::err_grad_u=-1.0d0  ! Velocity gradient
@@ -58,8 +63,14 @@ module hydro_parameters
   real(dp)::floor_d=1d-10     ! Density floor
   real(dp)::floor_u=1d-10     ! Velocity floor
   real(dp)::floor_p=1d-10     ! Pressure floor
-  ! TODO: allow for discontinuity-based refine on non-standard hydro vars?
   real(dp)::mass_sph=0.0d0     ! mass_sph
+  ! TODO allow for discontinuity-based refine on non-standard hydro vars:
+#if NENER>0
+  real(dp),dimension(1:NENER)::err_grad_prad=-1
+#endif
+#if NVAR>NHYDRO+NENER
+  real(dp),dimension(1:NVAR-NHYDRO-NENER)::err_grad_var=-1
+#endif
   real(dp),dimension(1:MAXLEVEL)::jeans_refine=-1
 
   ! Initial conditions hydro variables
--- a/hydro/hydro_parameters.f90
+++ b/hydro/hydro_parameters.f90
@@ -39,12 +39,8 @@ module hydro_parameters
   real(dp),dimension(1:MAXBOUND)::u_bound=0
   real(dp),dimension(1:MAXBOUND)::v_bound=0
   real(dp),dimension(1:MAXBOUND)::w_bound=0
-#if NENER>0
-  real(dp),dimension(1:MAXBOUND,1:NENER)::prad_bound=0
-#endif
-#if NVAR>NDIM+2+NENER
-  real(dp),dimension(1:MAXBOUND,1:NVAR-NDIM-2-NENER)::var_bound=0
-#endif
+  ! TODO allow for other variables in inflow?
+
   ! Refinement parameters for hydro
   real(dp)::err_grad_d=-1.0d0  ! Density gradient
   real(dp)::err_grad_u=-1.0d0  ! Velocity gradient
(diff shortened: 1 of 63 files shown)


mhd/hydro_parameters.f90
```diff
@@ -5,16 +5,30 @@ module hydro_parameters
   use amr_parameters
 
   ! Number of independant variables
+
+  ! Euler variables: density, velocity, pressure
+  ! always 3 velocities when MHD
+  integer,parameter::neul=5
```
nbrucy, line 12: Could neul be named neuler instead?

mhd/hydro_parameters.f90
```diff
@@ -5,16 +5,30 @@ module hydro_parameters
...
+#else
+  integer,parameter::nhydro=NHYDRO
+#endif
+  ! non-thermal energies
 #ifndef NENER
   integer,parameter::nener=0
 #else
   integer,parameter::nener=NENER
 #endif
+  ! total amount of variables
```
nbrucy, line 25: total amount of independent variables? (to distinguish from nvar_all)

mhd/hydro_parameters.f90
```diff
@@ -32,59 +46,47 @@ module hydro_parameters
...
-  real(dp)::floor_d=1d-10   ! Density floor
-  real(dp)::floor_u=1d-10   ! Velocity floor
-  real(dp)::floor_p=1d-10   ! Pressure floor
-  real(dp)::floor_A=1d-10   ! Bx floor
-  real(dp)::floor_B=1d-10   ! By floor
-  real(dp)::floor_C=1d-10   ! Bz floor
-  real(dp)::floor_b2=1d-10  ! B L2 norm floor
-  real(dp)::mass_sph=0.0D0   ! mass_sph
-#if NENER>0
-  real(dp),dimension(1:NENER)::err_grad_prad=-1.0
```
nbrucy, line 67: Where are these gone?

mhd/output_hydro.f90
```diff
@@ -110,41 +111,44 @@ subroutine backup_hydro(filename, filename_desc)
...
                  call generic_dump(field_name, info_var_count, xdp, unit_out, dump_info_flag, unit_info)
               end do
 #endif
-              do i = 1, ncache ! Write thermal pressure
+              ! Write thermal pressure
+              do i = 1, ncache
                  d = max(uold(ind_grid(i)+iskip, 1), smallr)
-                 u = uold(ind_grid(i)+iskip, 2)/d
-                 v = uold(ind_grid(i)+iskip, 3)/d
-                 w = uold(ind_grid(i)+iskip, 4)/d
```
nbrucy, line 125: Why are these changes relevant for the PR?

nbrucy (changes requested):
This makes the counting of variable easier to understand, with no consequences for end user (maybe a very small performance loss due to moving some computation at runtime?)

Maybe this can be further improved by adding some developer documentation in `doc/dev_docs` explaining what are the different number of variables `nvar`, `nhydro`, `neul` (why not `neuler` btw?), `nvar_all`, `nener`, etc ... and how they can be used.

mhd/hydro_parameters.f90
tinecolman: I used neul because it was already used in some places in the code. I simple extended its use.

nbrucy: Should we keep it like this?

mhd/output_hydro.f90
tinecolman: Here I've synchronized the hydro and mhd version of the file. Reducing the amount of duplicate code between hydro and mhd was one of the motivations for introducing these new variables.

hydro/hydro_parameters.f90
```diff
@@ -6,16 +6,28 @@ module hydro_parameters
...
   integer,parameter::nener=NENER
 #endif
+  ! total amount of variables
 #ifndef NVAR
-  integer,parameter::nvar=ndim+2+nener
+  integer,parameter::nvar=nhydro+nener
 #else
   integer,parameter::nvar=NVAR
 #endif
+  integer,parameter::nvar_all=nvar
```
cphyc, line 29: Regarding @nbrucy's comment about performance, all the variables are defined as compile-time, so the compiler should optimise them as such. If truly this is a worry, one could also replace these definitions with compiler directives, i.e. in the Makefile, create `DEFINES = -DNVAR_all=[...] -Dnhydro=[...]` etc. and used capitalized form to distinguish from runtime params.

mhd/hydro_parameters.f90
tinecolman: You mean err_grad_prad and err_grad_var? They are unused variables, so I removed them.

mhd/hydro_parameters.f90
tinecolman: In my opinion, neuler is not particulary more clear than neul. In any case, it is explained where the variable is defined. So I would opt for keeping neul.

mhd/hydro_parameters.f90
rteyssier: They are unused but they could be used in the future or in a patch.
I would rather keep them.

rteyssier (changes requested):
I would rather keep the radiation and passive scalar initial conditions and boundary conditions variables.

hydro/hydro_parameters.f90
tinecolman: For nhydro it is there -DNHYDRO=$(NHYDRO). Should I add the others? I would think it is not an issue, but we don't really have performance tests in place that would cover the different configurations of these parameters so I can't know for sure.

mhd/hydro_parameters.f90
tinecolman: I put them back.

mhd/hydro_parameters.f90
tinecolman: Both nvar and nvar_all are total number of variables in a way...

tinecolman:
I've added some dev doc. Can be the start of a more detailed documentation on some implementation aspects.

rteyssier (approved):

nbrucy (approved):
I added the new doc page in the index and corrected some typos (see https://ramses-organisation--36.org.readthedocs.build/en/36/). Good for me!

PR #42, opened 2024-11-01
Add doc build to tests

robelgeda:
This PR adds test to check if the docs build correctly

## PR Checklist

- [x] Pull request points to the `dev` branch.
- [ ] If it is a bugfix, the PR should have label `on-merge: backport-to-stable`.
- [ ] For new code features, the PR should have been tested (please include the results of the tests below or, when possible, add a new test for the feature).

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -69,7 +69,7 @@ jobs:
           uses: actions/setup-python@v5
           with:
             python-version: ${{ matrix.python-version }}
-            
+
         - name: Install RTD requirements
           run: pip install -r doc/rtd_requirements.txt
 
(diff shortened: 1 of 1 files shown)


robelgeda:
Ok, if tests pass here we can merge this before new docs are added and it will test them in the pr

cphyc (approved):

cphyc (approved):
Actually, it'd be great to include the doc as a build artifact so we can check them?

robelgeda:
I have added this feature, now it uploads the built docs after the test finishes (both the pdf and html). They can be downloaded as a zipped file! I will update the ticket with instructions.

nbrucy:
Thanks! That is great!

PR #44, opened 2024-11-04
Add fortran linter to autoformating

cphyc:
This adds an autoformatting step to all PRs using https://github.com/cphyc/fortran-linter (disclaimer, I am the main author).

This is an opiniated formatter _à la_ black for Python, which applies a set of rules onto the code to ensure it is formatted consistently.

## How does it work in the PR workflow?

In practice, this adds another test to the testing suite. Whenever a piece of code is touched, the autoformatter will be run on the corresponding file. If the file is properly formatted, nothing happens. If it isn't (e.g., the autoformatter would format it differently), then the "pre-commit" test will fail.

## Discussion point

Pre-commit ci can either:
1. Be run automatically from `pre-commit.ci`
	1. and be triggered manually to fix issues by commenting `pre-commit.ci run` on the PR,
    2. or automatically update each PR with the formatting fixed,
	3. or be triggered manually by adding the label `pre-commit.ci run` to the PR.
2. Be run automatically as a Github action and autofixes the PR

At the moment, option 2. is implemented. The positive thing is that it will only run on the files that have been changed. It might be possible to only trigger autofixing with a label, but I haven't found how yet.

## Examples
First 50 lines of `pm/star_formation.f90`:
Before
```fortran
#if NDIM==3
subroutine star_formation(ilevel)
  use amr_commons
  use pm_commons
  use hydro_commons
  use poisson_commons
  use cooling_module, ONLY: XH=>X
  use constants, only: Myr2sec, Gyr2sec, mH, pi, rhoc, twopi
  use random
  use mpi_mod
  implicit none
#ifndef WITHOUTMPI
  integer::info,info2,dummy_io
  integer,parameter::tag=1120
#endif
  integer::ilevel
  !----------------------------------------------------------------------
  ! Description: This subroutine spawns star-particle of constant mass
  ! using a Poisson probability law if some gas condition are fulfilled.
  ! It modifies hydrodynamic variables according to mass conservation
  ! and assumes an isothermal transformation...
  ! On exit, the gas velocity and sound speed are unchanged.
  ! New star particles are synchronized with other collisionless particles.
  ! Array flag2 is used as temporary work space.
  ! Yann Rasera  10/2002-01/2003
  !----------------------------------------------------------------------
  ! local constants
  real(dp)::d0,mgas,mcell
  real(dp)::scale_nH,scale_T2,scale_l,scale_d,scale_t,scale_v
  real(dp),dimension(1:twotondim,1:3)::xc
  ! other variables
  integer ::ncache,nnew,ivar,ngrid,icpu,index_star,ndebris_tot,ilun=10
  integer ::igrid,ix,iy,iz,ind,i,n,iskip,nx_loc,idim
  integer ::ntot,ntot_all,nstar_corrected,ncell
  logical ::ok_free
  real(dp)::d,x,y,z,u,v,w,e,tg,zg
  real(dp)::mstar,dstar,tstar,nISM,nCOM,phi_t,phi_x,theta,sigs,scrit,b_turb,zeta
  real(dp)::T2,nH,T_poly,cs2,cs2_poly,trel,t_dyn,t_ff,tdec,uvar
  real(dp)::ul,ur,fl,fr,trgv,alpha0
  real(dp)::sigma2,sigma2_comp,sigma2_sole,lapld,flong,ftot,pcomp=0.3d0
  real(dp)::divv,divv2,curlv,curlva,curlvb,curlvc,curlv2
  real(dp)::birth_epoch,factG
  real(kind=8)::mlost_all,mtot_all
#ifndef WITHOUTMPI
  real(kind=8)::mlost,mtot
#endif
  real(kind=8)::PoissMean
  real(dp),dimension(1:3)::skip_loc
  real(dp)::dx,dx_loc,scale,vol_loc,dx_min,vol_min,d1,d2,d3,d4,d5,d6
  real(dp)::mdebris
```
After
```fortran
#if NDIM == 3
subroutine star_formation(ilevel)
    use amr_commons
    use pm_commons
    use hydro_commons
    use poisson_commons
    use cooling_module, ONLY: XH => X
    use constants, only: Myr2sec, Gyr2sec, mH, pi, rhoc, twopi
    use random
    use mpi_mod
    implicit none
#ifndef WITHOUTMPI
    integer :: info, info2, dummy_io
    integer, parameter :: tag=1120
#endif
    integer :: ilevel
    ! ----------------------------------------------------------------------
    ! Description: This subroutine spawns star-particle of constant mass
    ! using a Poisson probability law if some gas condition are fulfilled.
    ! It modifies hydrodynamic variables according to mass conservation
    ! and assumes an isothermal transformation...
    ! On exit, the gas velocity and sound speed are unchanged.
    ! New star particles are synchronized with other collisionless particles.
    ! Array flag2 is used as temporary work space.
    ! Yann Rasera  10/2002-01/2003
    ! ----------------------------------------------------------------------
    ! local constants
    real(dp) :: d0, mgas, mcell
    real(dp) :: scale_nH, scale_T2, scale_l, scale_d, scale_t, scale_v
    real(dp), dimension(1:twotondim, 1:3) :: xc
    ! other variables
    integer :: ncache, nnew, ivar, ngrid, icpu, index_star, ndebris_tot, ilun=10
    integer :: igrid, ix, iy, iz, ind, i, n, iskip, nx_loc, idim
    integer :: ntot, ntot_all, nstar_corrected, ncell
    logical :: ok_free
    real(dp) :: d, x, y, z, u, v, w, e, tg, zg
    real(dp) :: mstar, dstar, tstar, nISM, nCOM, phi_t, phi_x, theta, sigs, scrit, b_turb, zeta
    real(dp) :: T2, nH, T_poly, cs2, cs2_poly, trel, t_dyn, t_ff, tdec, uvar
    real(dp) :: ul, ur, fl, fr, trgv, alpha0
    real(dp) :: sigma2, sigma2_comp, sigma2_sole, lapld, flong, ftot, pcomp=0.3d0
    real(dp) :: divv, divv2, curlv, curlva, curlvb, curlvc, curlv2
    real(dp) :: birth_epoch, factG
    real(kind=8) :: mlost_all, mtot_all
#ifndef WITHOUTMPI
    real(kind=8) :: mlost, mtot
#endif
    real(kind=8) :: PoissMean
    real(dp), dimension(1:3) :: skip_loc
    real(dp) :: dx, dx_loc, scale, vol_loc, dx_min, vol_min, d1, d2, d3, d4, d5, d6
    real(dp) :: mdebris
```


rteyssier:
Nice idea but this breaks the auto-formatting in emacs. So not a big fan...

cphyc:
How does it break auto-formatting in emacs? I am still confused by this statement.

Is it just the indentation? If so, that can be configured here.

rteyssier:
Yes the indentation is not the same as the emacs fortran-mode.

cphyc:
How many spaces is emacs' fortran-mode using? Three?

rteyssier:
I think 2 then 3.

rteyssier:
https://emacs.stackexchange.com/questions/63551/i-am-writing-fortran-code-in-emacs-but-the-tab-key-behaviour-changes-in-f90-mode￼
I am writing Fortran code in emacs but the tab key behaviour changes in F90 mode. Can I change it back?
emacs.stackexchange.com

cphyc:
It seems to be +2 spaces for subprogram/routines/functions/blocks and +3 otherwise.

Just tried in emacs:
```fortran
program test
  integer, parameter :: size = 4

  real(kind=8), dimension(size) :: my_array

  ! Two spaces
  if (.true.) then
     ! +3
     print*, "Interesting!"
     do
        ! +3
        block
          ! +2
        end block
     end do
  end if

  block foo
    ! Four...
  end block

contains
  subroutine subtest
    ! Four...
  end subroutine subtest
  
end program test

```

cphyc:
OK, so the issue is:
- fortran-linter as well as fprettify only support indenting with a constant amount of spaces (configurable)
- emacs indents with up to 7 (?!) different number of spaces, depending on context
- virtually all other text editors indent with constant spacing (including the one(s) I've been using, apart from emacs).

I understand this could be annoying, but should this change be blocked to support emacs, or should we propose an emacs config file to match what other editors are doing?

To be fair, I don't think the entirety of the code is consistently indented; for example the pieces I contributed have multiples of 2 or 3 for their indentation (because it's otherwise not supported by my editor).

rteyssier:
We have been living for years without automatic formatting so I would say we can keep it that way.

rteyssier:
Let see what other people think.

nbrucy:
I think the indentation is very inhomogeous, and it will be good to have all new code with a constant formatting (bad indentation makes the code very hard to understand).

Providing a config parameter for emacs can be useful, but maybe if the correction is automatic at the PR stage maybe people don't have to care about it.

nbrucy:
Currently the precommit fails on end of line without fixing them.

cphyc:
> Currently the precommit fails on end of line without fixing them.

I know - I won't fix it if we decide that we don't need all of this work :)

tinecolman:
So should we just decide to stick with the 2-3-3 spaces rule to homogenize the code? A weird rule is better than no rule.

cphyc:
The issue is that the 2-3-3 rule is hard to implement since indentation becomes context-dependent. As far as I know, it's only supported by emacs; VS Code doesn't support it and I know of no autoformatter that do.

I think we either drop the 2-3-3 rule (@rteyssier) or that PR.

rteyssier:
Let’s organize a vote.I am happy with whatever outcome but this is a good example of a decision that should be taken via majority rule.Romain On Jan 14, 2025, at 3:19 PM, Corentin Cadiou ***@***.***> wrote:﻿
The issue is that the 2-3-3 rule is hard to implement since indentation becomes context-dependent. As far as I know, it's only supported by emacs; VS Code doesn't support it and I know of no autoformatter that do.
I think we either drop the 2-3-3 rule ***@***.***) or that PR.

rteyssier (commented):
What should we do about this PR?
It has been there forever and no consensus has been reached.

tinecolman:
In the description of this PR there is no list of what the actual rules are, since those are implemented in a standalone package from Corentin. I'm not sure I like the idea that the rules are defined outside of ramses.

But I do agree that auto-formatting would be nice to have. If it is done at PR verification stage, the developer indeed does not need to pay attention to it themselves. They just use whatever they like best and the formater will change it to something which is uniform, and more easy to read for other developers.

However, how it is done in the PR is maybe not very democratic. Also, after a quick look at Corentin's package, it seems there are a lot of rules. Maybe we want to stick with some simple basic rules?
I would propose to hold a discussion session about formatting on the next RUM. We could start from subset of the rules that are implemented in Corentin's package. For each rule, we vote whether we want to apply it as is, apply a modified version of the rule, or not apply it.

tinecolman:
For example, I don't really like the rule 
> "A line should not exceed 120 characters (this is somehow already extreme). The maximum line length can be controlled from the CLI."

With modern IDEs this is obsolete no?  Also gcov doesn't like multi-line statements, so those made analysing the code coverage a bit annoying.

cphyc:
I agree - let's chat about this at RUM. For reference, there are currently 33 rules (but some are very similar, for example replacing `.le.` with `<` and `.ge.` with `>`). Voting by group might be a good idea if the concept of a linter is approved. It is relatively easy for me to modify the package to allow only a subset of rules to be applied.

Regarding line length, this is a matter of personal opinion. I don't like having to go fullscreen to be able to read a file because I often have two or three ones side-by-side, so I'm happy with ~100 max characters per line.

rteyssier (commented):
I suggest we close this pull request.
It has been around for almost a year with no clear decision whether we should use it.

PR #59, opened 2025-01-20
clean up unused functions in poisson multigrid

tinecolman:


poisson/multigrid_fine_commons.f90
```diff
@@ -1371,93 +1371,93 @@ end subroutine make_reverse_mg_dp
...
-      if(mod(myid,IOGROUPSIZE)/=0 .and.(myid.lt.ncpu))then
-         dummy_io=1
-         call MPI_SEND(dummy_io,1,MPI_INTEGER,myid-1+1,tag, &
-              & MPI_COMM_WORLD,info2)
-      end if
-   endif
-#endif
-
-end subroutine dump_mg_levels
+!subroutine dump_mg_levels(ilevel,idout)
```
nbrucy, line 1374: Maybe leave a comment to explain why it was not removed? 
```suggestion
!subroutine dump_mg_levels(ilevel,idout)
! commented out but kept for debugging (used in utils/py)
```

patch/hydro/planet/multigrid_coarse.f90
cphyc, line 1: Would that mean that the Makefile wouldn't work anymore by sampling adding `PATCH=hydro/planet`, since `multigrid_coarse.f90` won't be included?

cphyc (approved):
Thanks for the spring cleanup :)
Is there a reason to keep `multigrid_topdown.f90` at all, since it only contains functions that aren't used elsewhere and the file isn't included at all?

tinecolman:
> Thanks for the spring cleanup :)
> Is there a reason to keep `multigrid_topdown.f90` at all, since it only contains functions that aren't used elsewhere and the file isn't included at all?

It looked functional, so maybe someone will want to use it? It was not stated why it was obsolete, so I could not judge whether it was OK to remove it. I was mostly concerned with it ruining my coverage statistics.

patch/hydro/planet/multigrid_coarse.f90
tinecolman: Yes. But patches are not my concern. We  need a different strategy for keeping those updated.

PR #61, opened 2025-01-20
Add coverage statistics to test suite using gcov

tinecolman:
Added an option -s (for statistics) to run the test suite with gcov. Statistics will be gathered for each test separately and temporarily stored in the respective test directory. At the end, a script will combine the statistics from all tests into one folder called coverage. This folder contains a file for each source file that was compiled, taking care of duplicate file names. A report is printed to screen with the percentage of lines that was executed for each file, as well as the overall coverage (for source files compiled by the test suite, i.e. this currently excludes aton/).


tinecolman:
Known issues:
- multi-line statements are not always counted properly. To be cleaned up in another PR
- some gcov warnings are printed to screen in between test. I didn't find a way to get rid of those so far.
- a line may be wrongly marked as unexecutable if the variable name starts with one of the fortran keywords. For example 'endothermal' would be confused with the keyword 'end'.

tinecolman:
Screenshot from 2025-01-20 23-18-26
This is what it looks like.

nbrucy (approved):
Nice! Next step could be having this in the CI: https://docs.codecov.com/docs/quick-start

tests/hydro/barotrop/plot-barotrop.py
```diff
@@ -58,6 +58,6 @@
 fig.savefig('barotrop.pdf',bbox_inches='tight')
 
 # Check results against reference solution
-red_tol = 2.0e-12
+red_tol = 5.0e-12
```
rteyssier: Why is this changed?
Similar issue in another pull request...

rteyssier: @tinecolman why change the plot-barotrop.py file?

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+                if line.startswith("        -:    0:Source:"):
+                    source_file = line.split("Source:")[1].strip()
+                    continue
+
+                if source_file is None:
+                    continue  # Skip until the source file is identified
+
+                # Match GCOV execution counts: execution_count: line_number: line_content
+                match = re.match(r'^\s*(\-|#####|\d+\*?|\d+):\s*(\d+):\s*(.*)', line)
+                if match:
```
cphyc: For the sake of readability, I'd suggest:
```suggestion
                if not match:
                    continue
```
and deindent the rest :)

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+                # Identify source file
+                if line.startswith("        -:    0:Source:"):
+                    source_file = line.split("Source:")[1].strip()
+                    continue
+
+                if source_file is None:
+                    continue  # Skip until the source file is identified
+
+                # Match GCOV execution counts: execution_count: line_number: line_content
+                match = re.match(r'^\s*(\-|#####|\d+\*?|\d+):\s*(\d+):\s*(.*)', line)
```
cphyc: It'd be useful to have some explanation of what this beautiful regexp matches. As far as I could tell, this'd be:
```
     -: 12345: whatever
 #####:   0: also whatever
 9876:   1234:  still whatever
```

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+                if match:
+                    count, line_number, line_content = match.groups()
+                    # get the line of code, preserving spaces
+                    i_first_colon = line.find(':')
+                    i_second_colon = line[i_first_colon+1:].find(':')
+                    line_content = line[i_first_colon+i_second_colon+2:-1]
```
cphyc: Since this has been extracted above from the regexp, what's the point of all this? Can't it just be removed?

This chunk of code, as far as I understand, find the first ":", then how far ahead is the second, then extracts whatever is left on that line (but the last character, which is a `\n`). To me, that sounds *exactly* as `line_content`, as extracted from the regexp (except it contains the `\n`).

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+
+                    if (count == "-") or (count == "#####"):  # Line not executable or not executed
```
cphyc:
```suggestion
                    if count in ("-", "#####"):  # Line not executable or not executed
```

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+                                    "INTEGER", "LOGICAL", "CHARACTER", "REAL","COMPLEX", "TYPE",
+                                    "module",  "private", "public", "contains", "interface",
+                                    "MODULE",  "PRIVATE", "PUBLIC", "CONTAINS", "INTERFACE",
+                                    "use", "implicit", "namelist",
+                                    "USE", "IMPLICIT", "NAMELIST",
+                                    "else", "end", "case", "continue",
+                                    "ELSE", "END", "CASE", "CONTINUE",
+                                    "!", "#", "&",
+                                    "1","2","3","4","5","6","7","8","9") #format strings
+                        if len(stripped_line) == 0 or stripped_line.startswith(keywords):
```
cphyc:
```suggestion
                        stripped_line = line_content.strip()
                        keywords = (
                            'integer', 'logical', 'character', 'real', 'complex',
                            'type', 'module', 'private', 'public', 'contains',
                            'interface', 'use', 'implicit', 'namelist', 'else',
                            'end', 'case', 'continue', '!', '#', '&',
                            '1', '2', '3', '4', '5', '6', '7', '8', '9', # format strings
                        )
                        if len(stripped_line) == 0 or stripped_line.lower().startswith(keywords):
```

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+                        if isinstance(current_count, int):
+                           current_count += count
+                        else:
+                           current_count = count
+
+                    if line_content != current_line_content and current_line_content != "":
+                        print(f"Warning: Line content mismatch in {source_file} at line {line_number}")
+                        print(f"  Previous line content: {current_line_content}")
+                        print(f"  Current line content: {line_content}")
+                        print(f"  Using previous line content and count for aggregation.")
```
cphyc: This should go in the error not in stdout.
```suggestion
                        print(f"Warning: Line content mismatch in {source_file} at line {line_number}", file=sys.stderr)
                        print(f"  Previous line content: {current_line_content}", file=sys.stderr)
                        print(f"  Current line content: {line_content}", file=sys.stderr)
                        print("  Using previous line content and count for aggregation.", file=sys.stderr)
```
Note: the suggestion assumes `sys` has been imported

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+    def save_aggregated_coverage(self, output_directory):
+        """
+        Save the aggregated coverage data for each source file.
+        """
+        full_code_tot = 0
+        full_code_coverage = 0
+        os.makedirs(output_directory, exist_ok=True)
+
+        # sort files alphabetically
+        sorted_files = sorted(list(self.coverage_data.keys()))
```
cphyc:
```suggestion
        sorted_files = sorted(self.coverage_data.keys())
```
or even
```suggestion
        sorted_files = sorted(self.coverage_data)
```
but the latter is a bit more ambiguous (although equivalent)

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+
+        for source_file in sorted_files:
+            coverage = self.coverage_data[source_file]
+            num_lines_tot = 0
+            num_lines_exec = 0
+            # Use the source file name to create an output file, keeping info on the source directory
+            source_filename_parts = source_file.split('/')
+            output_name = (source_filename_parts[-2]).upper() + '_' + source_filename_parts[-1]
```
cphyc:
```suggestion
            *_head, parent_folder, fname = source_file.split('/')
            output_name = f"{parent_folder.upper()}_{fname}"
```

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+
+                    # gather stats for file
+                    if count!='-':
+                        num_lines_tot = num_lines_tot+1
+                        if isinstance(count, int) and count>0:
+                            num_lines_exec = num_lines_exec+1
+
+            # print stats on current file and add to total
+            if num_lines_tot>0:
+                print("{} {:6.2f}%   ({}/{})".format(source_file.ljust(50), 100*float(num_lines_exec)/num_lines_tot, num_lines_exec,num_lines_tot))
```
cphyc:
```suggestion
                coverage = 100 * num_lines_exec / num_lines_tot
                print(f"{source_file:50s} {coverage:6.2f}%   ({num_lines_exec}/{num_lines_tot})")
```

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+                full_code_tot = full_code_tot + num_lines_tot
+                full_code_coverage = full_code_coverage + num_lines_exec
+
+        print("Total code coverage {:.2f}% ({}/{})".format(100*float(full_code_coverage)/full_code_tot,full_code_coverage,full_code_tot))
```
cphyc:
```suggestion
        coverage = 100*full_code_coverage/full_code_tot
        print(f"Total code coverage {coverage:.2f}% ({full_code_coverage}/{full_code_tot})")
```

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
...
+# - in turb_force_utils, stop is identified as a branch
+
+import os
+import re
+from collections import defaultdict
+from glob import glob
+
+class GCovParser:
+    def __init__(self):
+        # Dictionary to store data as {source_file: {line_number: (execution_count, line_content)}}
```
cphyc: Isn't it instead?
```suggestion
        # Dictionary to store data as {source_file: {line_number: (execution_count, line_number, line_content)}}
```
If so, why `""` as a default line number rather than some integer (like -1)?

cphyc (approved):
Thanks, that's a super nice addition and I'm surprised it's so contained! I'm looking forward to having the possibility to estimate the actual coverage of the code!

Out of curiosity, is there a reason not to activate it by default in the test suite (apart from waiting to include in e.g. codedov)?

tinecolman:
> Out of curiosity, is there a reason not to activate it by default in the test suite (apart from waiting to include in e.g. codedov)?

There are two main reasons.
1) It's slow because of the instrumentation added by the coverage flags + you need to compile with -O0.
2) It generates a ton of extra temporary files. Because the way the test suite is set up, we need to keep a copy of each source code file for each test. When the code is recompiled, the counters are reset. Normally gcov would just add the statistics for each new run of the same executable.

tests/multi_gcov_aggregator.py
```diff
@@ -0,0 +1,145 @@
+# Script to merge coverage files generated by gcov
+# Author: Noe Brucy, Tine Colman
+# Known issues:
+# - in rt_spectra, gcov identified some variable declarations as executed.
+# - in turb_force_utils, stop is identified as a branch
+
+import os
+import re
```
tinecolman, line 8:
```suggestion
import re
import sys
```

tests/multi_gcov_aggregator.py
tinecolman: No, the 3th entry is a list of the tests in which the line is executed. For now I disabled printing this because the line gets very long for lines that are touched by many tests.

tests/multi_gcov_aggregator.py
tinecolman: I added this because I wasn't happy with how the regex magic removed indentation, which made the final cov files unreadable. I will replace the regex with this, so it is more understandable.

rteyssier (approved):

tests/hydro/barotrop/plot-barotrop.py
tinecolman: The increase of tolerance is needed to prevent this test from failing on different machines and when using different number of processes. The poisson multigrid has trouble converging. It seems to be a typical problem for collapsing sphere kind of setups.

PR #63, opened 2025-01-21
add cosmo test

tinecolman:
Tests dark matter particles in a cosmological box without AMR


tinecolman:
Screenshot from 2025-01-21 15-59-33
This is what it looks like in the report.

rteyssier:
I don't think it is a good idea to have these big initial condition files in the repo.
It is better to download them from outside as part of the test pipeline.

tinecolman:
> I don't think it is a good idea to have these big initial condition files in the repo. It is better to download them from outside as part of the test pipeline.

Because it takes too long to download the code? For the github file limits it's ok. What is the maximum file size you would add to the repo?

rteyssier:
No it is more because if you keep adding such big files, the repo will be too big.
Especially to download for users,

tinecolman:
Ok, Noé will add a thing to store the files on the SNO website and download them when needed.

tests/poisson/cosmo/before-test.sh
```diff
@@ -0,0 +1,10 @@
+if [ ! -d "cosmo_128" ]; then
+    echo "cosmo_128 directory does not exist, dowloading it..."
+    wget https://ramses.cnrs.fr/wp-content/uploads/2025/01/cosmo_128.zip
```
cphyc, line 3: If we assume we're pushing towards having more comprehensive tests, shouldn't we move this `wget` step to, say, a step in the Github action? This would allow managing cache to speed up the testing (should that be necessary) and would provide a single entry point to add additional test data?

tests/poisson/cosmo/cosmo.nml
```diff
@@ -0,0 +1,37 @@
+&RUN_PARAMS
+cosmo=.true.
+pic=.true.
+poisson=.true.
+hydro=.false.
+nremap=0
+nsubcycle=1,1,30*2
+nstepmax=100
```
cphyc, line 8: How much CPU time does this test take? Time on GitHub actions is somewhat limited (although I never hit the cap).

nbrucy: The whole test suite is about 35min, we have 2000 min per month.

tests/poisson/cosmo/before-test.sh
nbrucy: I think we want to keep the test runnable outside Github. That said, since there is a check whether the cosmo ic directory is there or not, I can also add it as a Github action step.

tests/poisson/cosmo/cosmo.nml
cphyc: Would it be worth to downgrade the resolution to, say, 64^3? It would make the test faster, and since we're testing accuracy down to $10^{-13}$, it should be equally stringent, shouldn't it?

If this is too much work on @tinecolman, I'm happy to just sh*t up and accept the PR :)

tests/poisson/cosmo/before-test.sh
cphyc: Sure, and my concern is exactly about this: if we end up with a few places that are downloading stuff from a public URL, it will become increasingly hard to set up a clean testing environment if everything is downloaded just-in-time.

Presently we do:
1. Run test suite
  - Run test 1
  - Run test 2
  - [...]
  - Run test n ... Ah, we're missing a file! Let's download it!

What I suggest is:
1. Download all files required for testing
2. Run test suite [assuming everything has been downloaded]

This will make the test suite easier to maintain while providing a way of downloading all files required for tests beforehand (say, you're going to travel and want to run tests while on the go, you now have an option to do so).
In addition, adding a `wget` command in the test means you can't run the tests offline.

Note that this mimics the "setup/teardown" approach of tests, where you separate concerns about what is _required_ for a test to be run from the test itself.

tests/run_test_suite.sh
```diff
@@ -286,7 +286,7 @@ for ((i=0;i<$ntests;i++)); do
 
    # Compile source
    echo "Compiling source" | tee -a $LOGFILE;
-   MAKESTRING="make EXEC=${EXECNAME} MPI=${MPI} ${FLAGS}";
+   MAKESTRING="make -j ${NCPU} EXEC=${EXECNAME} MPI=${MPI} ${FLAGS}";
```
cphyc: Is this safe with the current version of the Makefile? For ages, compiling with `-j` would result in some files trying to be compiled before their dependencies.

tests/poisson/cosmo/before-test.sh
nbrucy: With the last commit we do :
1. Download all files required for testing
2. Run test suite [if the file is missing we redownload it]

tests/run_test_suite.sh
nbrucy: hum ... indeed I forgot about that ...

nbrucy: and indeed it fails. I revert this.

tests/poisson/cosmo/cosmo.nml
tinecolman: I didn't have good 64^3 ICs laying around and since the 128^3 took only 1m3s on my laptop, I just went with that.

cphyc: If we want to, here are three sets of ICs at resolution 16, 32 and 64 (respectively) :)
[cosmo_16.zip](https://github.com/user-attachments/files/18507486/cosmo_16.zip)
[cosmo_32.zip](https://github.com/user-attachments/files/18507488/cosmo_32.zip)
[cosmo_64.zip](https://github.com/user-attachments/files/18507492/cosmo_64.zip)

cphyc (commented):
Another remark is that, before merging, we need to rewrite the git history to remove the binary files (which are still in commit d93758dfe014dbf4669e54813df9103d46309592, and thus still use space).

nbrucy:
> Another remark is that, before merging, we need to rewrite the git history to remove the binary files (which are still in commit [d93758d](https://github.com/ramses-organisation/ramses/commit/d93758dfe014dbf4669e54813df9103d46309592), and thus still use space).

That's true! Just did it.

tests/poisson/cosmo/before-test.sh
tinecolman: Github should always download the thing, no?
And if you run it on your laptop you should only download it if you run the cosmo test and don't already have the ICs in your directory.
So I think it's fine now?

tests/poisson/cosmo/cosmo.nml
tinecolman: The ones I used only have delta and velocity. Can I remove the other files?

rteyssier (approved):

tests/poisson/cosmo/cosmo.nml
cphyc: I think you can use them as is; they should work out of the box (famous last word)

cphyc (approved):

PR #66, opened 2025-01-22
split tests in ci

nbrucy:
This is a proposal to have separate CI steps for the different group of tests (has pros and cons).

## Pros
- easier to see which test are failing and which not
- good starting point for "on click" tests?
- see which tests take more time
- get individual test results faster (reduce the time we have to wait for tests results)

## Cons
- consume more Github Actions time (because the common part of downloading the code, etc.. is repeated)
- can't produce or global PDF summary (but a pdf per test family)

## PR Checklist

- [x] Pull request points to the `dev` branch
- [x] The pull request has a description
- [ ] If it is a bugfix that should be urgently merged to stable, the PR should have label `on-merge: backport-to-stable`.


cphyc (approved):
That looks good to me and seems pretty useful!

tinecolman:
Are we happy with the current test families?

tinecolman:
What is the impact on potentially adding automatic test coverage statistics'

nbrucy:
> Are we happy with the current test families?

For now I took what was already there and that looks fine. This can be changed afterwards but the [.github/workflows/build-test.yaml](https://github.com/ramses-organisation/ramses/pull/66/files#diff-bd4e577d8cce18da4cecc900fca127f9a0d57d706c9cdb3507b778c1d574a54f) file needs to be updated accordingly.

Maybe I should add this to the "Tests" page in the doc?

nbrucy:
> What is the impact on potentially adding automatic test coverage statistics'

I think codecov can deal with that ... But I would need to configure it to be sure.

PR #70, opened 2025-02-12
Remove uplmde

tinecolman:
This PR removes Collela's PLMDE hydro scheme, which previously could be activated using the namelist parameter `scheme='plmde'`. This scheme only worked in 1D, it does not work with AMR in 2D or 3D. 

>[!WARNING]
>scheme='plmde' is no longer supported.

Notes:
* The namelist parameter scheme is kept because it is used for MHD.
* Documentation has been updated.
* Patches have been updated.


doc/wiki/Hydro_Parameters.md
```diff
@@ -15,11 +15,10 @@ describe them now
...
 | `niter_riemann=10`    |  `Integer` | Maximum number of iterations used in the exact Riemann solver |
 | `slope_type=1` |  `Integer`    | Type of slope limiter used in the Godunov scheme for the piecewise linear reconstruction: slope_type=0: First order scheme, slope_type=1: MinMod limiter, slope_type=2: MonCen limiter, slope_type=3: Multi-dimensional MonCen limiter. In 1D runs only, it is also possible to choose:slope_type=4: Superbee limiter, slope_type=5: Ultrabee limiter |
 | `slope_mag_type=1` |  `Integer`    | Type of slope limiter used in the Godunov scheme in the MHD solver |
 | `pressure_fix=.false.` |  `logical`    | Activate hybrid scheme (conservative or primitive) for high-Mach flows. Useful to prevent negative temperatures. |
 | `beta_fix=0d0` |  `Real`    | With pressure_fix=.true., changes the threshold at which the energy is truncated |
-| `difmag=0d0` |  `Real`    | Modifies diffusive flux in the PLDME integrator (hydro only). |
 | `eta_mag=0d0` |  `Real`    | Modifies dtdiff in PLDME integrator (MHD only, warning, divide by zero error if eta_mag=0d0) |
```
nbrucy, line 24: Should'nt this be also removed? or renamed?

mhd/hydro_parameters.f90
```diff
@@ -110,11 +110,10 @@ module hydro_parameters
   real(dp)::gamma=1.4d0
   real(dp),dimension(1:512)::gamma_rad=1.33333333334d0
   real(dp)::courant_factor=0.5d0
-  real(dp)::difmag=0
   real(dp)::smallc=1.0d-10
   real(dp)::smallr=1.0d-10
   real(dp)::eta_mag=0
```
nbrucy, line 115: eta_mag - linked to PMLDE?

doc/wiki/Hydro_Parameters.md
nbrucy: Also you may want to add a more obvious depreciation warning here (outside the table):

```
:::{deprecated} 2025
The Collela's PLMDE schemehas been removed in 2025 ([PR #70](https://github.com/ramses-organisation/ramses/pull/70)
:::
```

nbrucy (changes requested):

doc/wiki/Hydro_Parameters.md
tinecolman: You mean eta_mag? It is used in the diffusion subroutine, which is called separately in amr_step. I think this is something different than PLDME?

nbrucy: Yes `eta_mag`. The docs says it's linked to PMLDE, maybe that part is wrong.

rteyssier (commented):
I think we should keep it.
This scheme is similar to Corner Transport Upwind by Collela.
Someone might want to use it.
Routine consup is also important.
I vote to reject this pull request.

tinecolman:
So it actually does work in 3D? But not with MHD? Instead of removing it, we should add a test then. Is there any specific case in which this scheme is favored over muscl or do I just use a sedov test?

rteyssier (changes requested):
I wold rather keep these files even if they are not used by the code but someone might use these functions.

PR #78, opened 2025-03-04
Some love to classical tracers

nbrucy:
This PR reallow loading classical tracers from ascii file + add a test with an example.

image

(should improve the coverage ;) )

## PR Checklist

- [x] Pull request points to the `dev` branch
- [x] The pull request has a description
- [x] If it is a bugfix that should be urgently merged to stable, the PR should have label `on-merge: backport-to-stable`.


nbrucy:
It's still a draft since I would like to improve the test and maybe improve the doc in the process.
Help appreciated.

tests/tracer/sedov-classical-tracers/sedov-classical-tracers.nml
```diff
@@ -0,0 +1,60 @@
...
+&REFINE_PARAMS
+err_grad_d=0.01
+interpol_type=1
+/
+
+
+&tracer_params
+MC_tracer = .false.
+tracer_feed_fmt = 'ascii'
+tracer_feed= 'tracers.txt'
```
cphyc, line 58: Is the corresponding file not missing?

nbrucy: The file is generated by the pre test hook  [before-test.sh](https://github.com/ramses-organisation/ramses/pull/78/files#diff-30b9011aa228889ba6ab827dd06252b24aa7cf5a296de080b162170604bc3aa7).

(the test is failing only because of small differences between my machine and the github runner).

nbrucy:
I updated the reference with a parallel run instead of single cpu so the test  pass, but I'm not at ease with the $10^{-4}$ variations with MPI on or off ...

pm/init_tracer.f90
```diff
@@ -115,8 +115,8 @@ subroutine load_tracers
                 write(*,*)'npartmax should be greater than',ipart, 'got', npartmax
                 stop
              endif
-             xp(ipart,:)  = xx(i,:)
-             vp(ipart,:)  = vv(i,:)
+             xp(ipart,:ndim)  = xx(i,:ndim)
+             vp(ipart,:ndim)  = vv(i,:ndim)
```
tinecolman, line 119: Is this going to be ok with mhd? Also if you want to write it like this, maybe put 1:ndim? I think this way of writing is more common in ramses.

nbrucy: Why not? This has nothing to do with MHD (and gas variable in general).

tinecolman: I was just wondering how xp and vp are allocated in 2D with MHD, since for hydro 3 dimension are needed. But indeed ndim is set independently and the particles don't care, so it's fine.

rteyssier (approved):

cphyc:
I could suggest another test - let me know if you think it's overkill.

The test would be simple: a box moving uniformly with velocity (1, 2, 3). The density as traced by the tracers should match (almost) exactly that of the gas, except for shot noise.

Another point is that, for tracers, we may want to test them with 2 MPI processes. This is a bit pointless for MC tracer because the random seeds depend on the number of processes.

nbrucy:
> I could suggest another test - let me know if you think it's overkill.
> 
> The test would be simple: a box moving uniformly with velocity (1, 2, 3). The density as traced by the tracers should match (almost) exactly that of the gas, except for shot noise.

I actually I wrote (almost) this test, with a constant gravitational force. I'll add it.
I'll also add a other way to load classical inplace (1 tracer at the center of each leaf cell).


> Another point is that, for tracers, we may want to test them with 2 MPI processes. This is a bit pointless for MC tracer because the random seeds depend on the number of processes.

By default, all the tests in the CI are run with 2 MPI processes.

nbrucy:
I'll merge this one and do a new PR for the  constant gravitational force test and the inplace loading.

PR #86, opened 2025-03-14
Add test to cover passive scalar code

tinecolman:
Screenshot from 2025-03-14 14-50-55

Description of the test:
A diamond shape is placed in a lower density medium which is moving to the left. The passive scalar is set to be 1 in the diamond shape and -1 in the background medium. Turbulence is generated in the wake at the corners of the diamond, 
causing the passive scalar to mix.

Reference solution generated using 2 mpi processes.
Execution time on 4 cpus is about 40 seconds on my machine.
I needed to increase the tolerance because the chaotic nature of the problem makes the solution sensitive to the number of processes used.


tests/hydro/mixing-scalar/plot-mixing-scalar.py
```diff
@@ -0,0 +1,57 @@
...
+
+fig, ax = plt.subplots(nrows=1, ncols=2, figsize=(8, 5))
+
+# Load RAMSES output
+data = visu_ramses.load_snapshot(2)
+x      = data["data"]["x"]
+y      = data["data"]["y"]
+dx     = data["data"]["dx"]
+rho    = data["data"]["density"]
+p      = data["data"]["pressure"]
```
nbrucy, line 17: `p` is not used

nbrucy (approved):

rteyssier (approved):

Issue #89, opened 2025-03-24
CI/CD tests shouldn't run if PR is not up-to-date with dev


tinecolman:
Merge first, then run tests otherwise they will be relaunched after the merge commit (running them twice).

nbrucy:
solved (for now) by #91

PR #91, opened 2025-03-25
Separate build and test workflow

nbrucy:
Make test manually triggered only, as a solution for #89 .
The idea is that the tests are now only run when the `run tests` label is added to the PR (or manually triggered in the branch). To rerun the test, remove and readd the label.


nbrucy:
(that was a test - maybe that generate to many notifications ...)

tinecolman:
I propose to try like this for some time. If we don't like it we can revert it.

tinecolman (approved):

PR #92, opened 2025-03-27
add basic test cases for nener

tinecolman:
- Add 1D sod-tube test with 2 non-thermal pressures, for the hydro and mhd solvers.
- Correct nener variable name for hydro file descriptor.
- Fix numpy issue in visualisation script.

>[!WARNING]
> The nener variable name `non_thermal_energy` was corrected into `non_thermal_pressure` in the output file.
> Please update your analysis script if needed.

Reference solution generated on 2 cpus.

Coauthor: Benoît Commerçon

Screenshot from 2025-03-27 10-39-45


hydro/output_hydro.f90
```diff
@@ -105,7 +105,7 @@ subroutine backup_hydro(filename, filename_desc)
                  do i = 1, ncache
                     xdp(i) = (gamma_rad(ivar-nhydro)-1d0)*uold(ind_grid(i)+iskip, ivar)
                  end do
-                 write(field_name, '("non_thermal_energy_", i0.2)') ivar-3
+                 write(field_name, '("non_thermal_pressure_", i0.2)') ivar-3
```
nbrucy, line 108: Can you explain why this was needed? seems to be potentially breaking for analysis script.

tinecolman: Because what is outputted is pressure, not energy. I don't think many people use NENER, so it should't break too many scripts.

nbrucy (approved):

PR #98, opened 2025-04-07
More modular read_params

tinecolman:
Refactor read_params to be more modular.
* Separate subroutines for each namelist block.
* Each dedicated subroutine defines their namelist block, reads the parameters, checks for errors, and further processes the input if needed. 
* Move calls to reading of hydro, sink etc namelists to follow the others.

Additionally:
* Initialize ncpu and myid only once, in the very beginning.
* Rename the local variable levelmax in unbinding to avoid confusion with levelmax from the namelist.

### DIFF
--- a/amr/amr_commons.f90
+++ b/amr/amr_commons.f90
@@ -31,7 +31,7 @@ module amr_commons
   CHARACTER(LEN=300)::namelist_file
 
   ! MPI variables
-  integer::ncpu,ndomain,myid,ierr,overload=1
+  integer::ncpu,ndomain,myid,overload=1
 
   ! Friedman model variables
   integer::n_frw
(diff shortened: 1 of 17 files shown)


nbrucy:
> Separate subroutines for each namelist block.

I'm afraid this will generate a lot of merge conflicts with personal versions that add new parameter ... Not sure that outweights the benefit ...

tinecolman:
> > Separate subroutines for each namelist block.
> 
> I'm afraid this will generate a lot of merge conflicts with personal versions that add new parameter ... Not sure that outweights the benefit ...

There are no dependencies on other files, so if they just replace the file, it should be fine. Also, it is straightforward to just move your new parameters to the correct function.

rteyssier (approved):

PR #108, opened 2025-05-05
[SPACE] Optimize and document nbor utils

tinecolman:
**Refactor routines for obtaining the 3cube neighbors**
* The routine get3cubefather is split into its two parts. Before, get3cubefather returned both the _cells_ and the _grids_, but only one of them was actually used after the call. Now, get3cubefather only returns the neighbor _cells_. A new subroutine get3cubefather_grids, returns the grids (only needed for poisson multigrid). This avoid unnecessary computations and memory access. 
* Move the first part of the routine get3cubepos to its own routine, named get_grids_of_nbor_cells, so it can be used by get3cubefather_grids.
* The input ind for get3cubepos and get_grids_of_nbor_cells is now an array of length nvector instead of an integer. This allows to remove the if condition inside the loop over cells to select the cells that sit at position ind, which hinders vectorization and creates more loop iterations:
```
        iok=0
        do i=1,ncell
           if(pos(i)==ind)then
```
* Use merge instead of if-else in get3cubepos, in an attempt to improve vectorization.

These changes reduce the cost of get3cubefather itself to almost nothing. The computational cost is now purely in the subroutines called by get3cubefather, rather than the loops at the end of the routine which were inefficient.
I've measured **performance gains** for several benchmarks:
* cosmo on uniform grid: marenostrum intel 6.2%, meluxina gnu 10.5%
* cosmo with amr:  marenostrum 8.2%, meluxina 11.2%
* sedov on uniform grid: 1-2%

**Documentation**
* Add an overview of the routines at the top of the file.
* Add more detailed descriptions of what each routine is doing.
* Try to make variable naming more intuitive/consistent.

This development was done in the context of the SPACE CoE.

### DIFF
--- a/amr/nbors_utils.f90
+++ b/amr/nbors_utils.f90
@@ -1,26 +1,44 @@
+! This file contains subroutines to obtains neighboring cells 
+! or grids. There are several options, depending on your needs:
+!   * get3cubefather: give the 
+!
+!
+!
+!   * getnborcells: obtain the 2**ndim direct neighboring cells
+!                   of a cell at position ind in a grid.
+!                   The result contains only the neighbors, not
+!                   the cell itself.
+!                   The input for this function is the output of
+!                   getnborgrids and ind.
+
+!   * getnborgrids: obtain the 2**ndim direct neighboring grids
+!                   of a grid. The result includes the grid itself
+!                   at position 0 in the array.
+!                   Often called before getnborcells.
+! 
 !##############################################################
 !##############################################################
 !##############################################################
 !##############################################################
-subroutine get3cubefather(ind_cell_father,nbors_father_cells,ncell,ilevel)
-  use amr_commons
+subroutine get3cubefather(ind_cell,nbor_cells,ncell,ilevel)
+  use amr_parameters, only:nx,ny,nz,ndim,ngridmax,nvector,threetondim
+  use amr_commons, only:ncoarse
   implicit none
   integer,intent(in)::ncell,ilevel
-  integer,dimension(1:nvector),intent(in)::ind_cell_father  ! input cell
-  !3x3x3 neighbors of input cell (including the center input cell)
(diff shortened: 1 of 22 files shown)


tinecolman:
Can someone go through the documentation I wrote to check if it is correct and clear? It took me some time to understand each of these routines in detail.

tinecolman:
To make clearer what has been changed:

Screenshot from 2025-05-12 18-24-00
Screenshot from 2025-05-12 18-24-31

amr/nbors_utils.f90
```diff
@@ -93,12 +123,63 @@ subroutine get3cubefather(ind_cell_father,nbors_father_cells,&
...
      end do
 
-     i1min=0; i1max=0
-     if(ndim > 0)i1max=1
+  else    ! else, more complicated...
+
+     ! Get the cell's position in its grid, that is the
+     ! index ind, between 1 and twotondim.
+     do i=1,ncell
+        pos(i)=(ind_cell(i)-ncoarse-1)/ngridmax+1  !integer devision
```
bcommerc, line 131: devision -> division

poisson/multigrid_fine_commons.f90
```diff
@@ -456,8 +456,8 @@ subroutine build_parent_comms_mg(active_loc, ifinelevel)
          ind_cell_father(i)=father( active_loc%igrid(istart+i-1) )
       end do
 
-      ! Compute neighbouring father cells and grids
-      call get3cubefather(ind_cell_father,nbors_father_cells, &
+      ! Compute neighbouring father grids
+      call get3cubefather_grids(ind_cell_father, &
```
rteyssier, line 460: This is weird. You store the nbors_father_cells in nbors_father_grids?

rteyssier (approved):

bcommerc (approved):

PR #110, opened 2025-05-16
[tests] Add a -r option to test restart

nbrucy:
Adds  an option to interject a restart in all tests. The option is NOT activated by default on the CI.

Some tests are failing because RAMSES is not binary exact. A solution would be to store conservative variable instead of primitive ones (as done in RAMSES ism) (see #123). 
There is a more serious issue with the MC tracer test (see #121)


tests/run_with_restart.py
```diff
@@ -0,0 +1,120 @@
...
+
+    nml = f90nml.read(nml_path)
+
+    if "tout" in nml["output_params"]:
+        tout = nml["output_params"]["tout"]
+        nml["output_params"]["tout"] = tout / 2
+    elif "aout" in nml["output_params"]:
+        aout = nml["output_params"]["aout"]
+        nml["output_params"]["aout"] = aout / 2
+    else:
```
tinecolman, line 64: You could add another case for when tend is used instead of tout?

tests/run_with_restart.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if "tout" in nml["output_params"]:
+        tout = nml["output_params"]["tout"]
+        nml["output_params"]["tout"] = [tout, 2*tout]
+    elif "aout" in nml["output_params"]:
+        aout = nml["output_params"]["aout"]
+        nml["output_params"]["aout"] = [aout, 2*aout]
+    else:
+        print("ERROR: tout or aout not found in output_params")
+        return
```
tinecolman: You could make this part a function to call with a factor parameter which is set to 2 or 1/2?

nbrucy:
The tests are working locally but silently failing on the CI ... Investigating ...

tests/sink/smbh-bondi/smbh-bondi.nml
```diff
@@ -34,8 +34,8 @@ boxlen=1. ! kpc
 /
 
 &OUTPUT_PARAMS
-foutput=1
```
tinecolman: This test is checking the accretion rate over multiple outputs to make the figure. Removing foutput makes this test fail because it does not find the other outputs (same for levelhack, since that uses the same setup). Maybe a solution could be to restart from the last output instead of output 2
```
nrestart="$(find . -maxdepth 1 -name output_'?????' | sort -n | tail -c 6)"
./ramses3d run.nml ${nrestart}
```

nbrucy: Yes I noticed. I'll fix that.

nbrucy: What I'll do is supporting another way to write output (foutput =1 and tend) so that I can halve tend and indeed restart from the last output.

tests/run_with_restart.py
nbrucy: Done

tests/run_with_restart.py
nbrucy: Done

nbrucy:
@cphyc There is something fishy with the MC tracer sedov test ... It crashes at restart after a "particle escaped allowed boundaries error". 
I'm investigating but maybe you have an idea on why this is happening. 

[test_results.pdf](https://github.com/user-attachments/files/20588695/test_results.pdf)

tinecolman (approved):

PR #111, opened 2025-05-20
simplify save_phi_old to remove allocate inside loop

tinecolman:
Rewrite save_phi_old so that the grid index is used right away instead of stored in an array which is allocated inside the loop over cpus. Allocate/deallocate is a slow operation and inside a loop can lead to memory performance issues. Here especially it could become expensive when ncpu is large.

### DIFF
--- a/poisson/interpol_phi.f90
+++ b/poisson/interpol_phi.f90
@@ -111,7 +111,7 @@ subroutine save_phi_old(ilevel)
            phi_old(igrid+iskip)=phi(igrid+iskip)
         end do
         igrid=next(igrid)
-      end do 
+      end do
   end do
 
 end subroutine save_phi_old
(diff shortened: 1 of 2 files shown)


poisson/interpol_phi.f90
```diff
@@ -102,24 +102,16 @@ subroutine save_phi_old(ilevel)
...
-        end do
-        ! Loop over cells
+     ! Loop over grids by following the linked list
+     igrid=istart
+     do i=1,ncache
+        ! Loop over cells in the current grid
         do ind=1,twotondim
            iskip=ncoarse+(ind-1)*ngridmax
-           ! save phi
-           do i=1,ncache
```
rteyssier, line 117: Loop is not vectorized anymore. Might affect performance.

PR #117, opened 2025-06-04
Remove hard cast of idp to short int

### LINKED ISSUE #116: Bug: Idp gets hard casted to short ints, even when compiling with -DLONGINT
Hi all,

I have been tracking a bug for the last couple of days when using long ints to store particle identities. 

I am compiling with `-DLONGINT`, but still seeing my `idp` values gets truncated to the maximum values of a short int (e.g. test case below on a simple dark matter only where ids get negative and are very close to the min and max value of an int32, pointing to a cast/overflow issue)
```
In [6]: s = pynbody.load("./output_00010/")

In [7]: s.d['iord'].min()
Out[7]: SimArray(-2147224578, dtype=int64)

In [8]: s.d['iord'].max()
Out[8]: SimArray(2144210113, dtype=int64)
```

I tracked it down to this line in the tree communication
https://github.com/ramses-organisation/ramses/blob/3081d6a2d4bf9b234c62ce0712052a8766476023/pm/particle_tree.f90#L1270

which hard casts `idp` to short ints irrespective of the length of `integer(i8b)` (the type of `communicator%fp`: see https://github.com/ramses-organisation/ramses/blob/3081d6a2d4bf9b234c62ce0712052a8766476023/amr/amr_commons.f90#L105-L118).

This likely went unnoticed for years because it wouldn't cause any crashes, just the garbling of IDs. It also wouldn't be an issue if compiling with `-DLONGINT` but using IDs that are less than 2,147,483,647.  

I am proposing a fix in PR #117, simply removing the cast as both `idp` and `communicator%fp` should both be `i8b` and not need any casting. The PR solves my issue and produces consistent IDs. Do let me know if you think this change would cause other issues elsewhere. 

@dchapon: I think the bug is also there in the LIGHT_COMMUNICATOR setup, although I have never tried this feature. Should the cast also be removed in the line 
https://github.com/ramses-organisation/ramses/blame/3081d6a2d4bf9b234c62ce0712052a8766476023/pm/particle_tree.f90#L1266 ?

Martin-Rey:
See #116

### DIFF
--- a/pm/particle_tree.f90
+++ b/pm/particle_tree.f90
@@ -1263,11 +1263,11 @@ subroutine empty_comm(ind_com,np,ilevel,icpu)
   do i=1,np
 #ifdef LIGHT_MPI_COMM
      levelp(ind_part(i))=int(emission_part(ilevel)%f8(2, offset_np+ind_com(i)-1), 4)
-     idp   (ind_part(i))=int(emission_part(ilevel)%f8(3, offset_np+ind_com(i)-1))
+     idp   (ind_part(i))=emission_part(ilevel)%f8(3, offset_np+ind_com(i)-1)
      typep(ind_part(i)) =int2part(int(emission_part(ilevel)%f8(4, offset_np+ind_com(i)-1), 4))
 #else
      levelp(ind_part(i))=int(emission(icpu,ilevel)%fp(ind_com(i),2), 4)
-     idp   (ind_part(i))= emission(icpu,ilevel)%fp(ind_com(i),3)
+     idp   (ind_part(i))=emission(icpu,ilevel)%fp(ind_com(i),3)
      typep(ind_part(i)) =int2part(int(emission(icpu,ilevel)%fp(ind_com(i),4), 4))
 #endif
   end do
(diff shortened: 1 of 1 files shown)


cphyc:
Looking at the files, it seems like the following line should be modified as well  https://github.com/ramses-organisation/ramses/blob/3081d6a2d4bf9b234c62ce0712052a8766476023/pm/star_formation.f90#L709 because `index_star` might (but in 99.99999% of the cases won't) overflow 2,147,483,647 ($2^{31}-1$).

Also as a quick sidenote, most of the tools in `utils/` won't work with `LONGINT`.

tinecolman (changes requested):
Could you apply the fix also to the light MPI version?

Martin-Rey:
> Looking at the files, it seems like the following line should be modified as well
> 
> https://github.com/ramses-organisation/ramses/blob/3081d6a2d4bf9b234c62ce0712052a8766476023/pm/star_formation.f90#L709
> 
> because `index_star` might (but in 99.99999% of the cases won't) overflow 2,147,483,647 

I agree with this but isn't it extremely unlikely? The total number of particles (stars+DM+other families) in the simulations would have to exceed 2 billions for this to occur, in which case we will hit a whole bunch of other issues because `npart`, `ntot`, `star_tot` are all implicitly assumed  to fit on 32 ints. 

It would be a natural extension of the `LONGINT` feature though and something to think about for the future. I currently have Ramses simulations with 600-700 millions particles, so we will hit this limitation sooner or later.

Martin-Rey:
@tinecolman That should do it!

tinecolman (approved):

nbrucy (approved):

PR #129, opened 2025-06-12
[tests] Check SOD tests against analytical solution

nbrucy:
This PR modifies the sod tube tests (hydro and ideal mhd)  to check that the error to the analytical solution does not change


tests/hydro/sod-tube/sod-tube-ref.dat
tinecolman, line 1: Defining the reference for these new variables as 0 is not consistent with the meaning of the columns in the table. Could you put the actual current deviation from the analytical solution as reference, to be consistent? Then we can easily see if the accuracy changes if we make a change to the code, which could be a decision criterion to update the reference solution or not.

tests/mhd/imhd-tube/imhd-tube-ref.dat
```diff
@@ -6,17 +6,25 @@ B_z_left : 4.5872245532670732e+01
...
 unit_d : 1.0000000000000000e+00
 unit_l : 1.0000000000000000e+00
 unit_t : 1.0000000000000000e+00
+velocity_avg_error : 0.
+velocity_med_error : 0.
 velocity_x : 1.9389631695548837e+02
-velocity_y : 4.0503914358611451e+02
-velocity_z : 4.4181525014765903e+01
+velocity_y : 4.0503914358611416e+02
+velocity_z : 4.4181525014765910e+01
```
tinecolman: You should not overwrite the solution for the velocity.

tinecolman (changes requested):

tests/hydro/sod-tube/sod-tube-ref.dat
nbrucy: done

tests/mhd/imhd-tube/imhd-tube-ref.dat
nbrucy: corrected

tinecolman (approved):

PR #132, opened 2025-06-12
[tests] Added a check for the clumpfinder

pocvirk:
Added a test for the clumpfinder to the poisson/cosmo test:
- edited cosmo.nml 
   - added CLUMPFIND_PARAMS block
   - added read methods for the clumpfinder output
   - added a reference solution to cosmo-ref.dat
   - added a specific clump-ref.txt file with the reference catalog (not used in the test but may be useful nonetheless)


tests/poisson/cosmo/plot-cosmo.py
tinecolman, line 1: The python script could use a bit of cleaning. 
- Could you remove the import statements which are not needed?
- Could you put all the things that need executing at the bottom, without tf __name__ == "__main__":
- Maybe define the functions at the top instead of between the executed code
- Would it be possible to read the file with numpy or so instead of pandas to avoid introducing another python package dependency for the CI
- could you remove the duplicate call to check_solution? It needs to be done only at the end.

tinecolman (changes requested):
An after-test.sh script should be added that removes the concatenated clump txt.

tinecolman (approved):

PR #133, opened 2025-06-12
[tests] RT dirac test

matthiasglz:
This PR adds a RT dirac test, originally from ramses-ism.
The final output time was lowered to reduce the execution time.

Remark that the reference solution is not the same as in ramses-ism.


rteyssier (approved):

rteyssier:
@nbrucy @matthiasglz hi guys,
what do we do about this pull request?

tinecolman:
> @nbrucy @matthiasglz hi guys, what do we do about this pull request?

This PR is still in draft. We ported this test from ramses-ism during the test suite hackaton, but for some reason the results are different. We haven't finished investigating why.

tinecolman:
> > @nbrucy @matthiasglz hi guys, what do we do about this pull request?
> 
> This PR is still in draft. We ported this test from ramses-ism during the test suite hackaton, but for some reason the results are different. We haven't finished investigating why.

It took too much time to figure out what the problem is, so I've decided to merge it as is.

PR #141, opened 2025-06-17
[tests] Do a parameter study in the sod-tube analytical test

nbrucy:
This should improve the coverage.

The idea is that the error to the analytical solution is computed for all the combinations available in `ana_parameters.py`


tinecolman:
It's complaining:
FileNotFoundError: [Errno 2] No such file or directory: 'sod-tube-parameter-study.npz'

nbrucy:
> It's complaining: FileNotFoundError: [Errno 2] No such file or directory: 'sod-tube-parameter-study.npz'

No other errors? 

Maybe a missing dependency?

`pip install scikit-learn`

tinecolman (approved):

PR #147, opened 2025-06-21
Directly store the flux in cmpflxm instead of using intermediate arrays fx and tx

tinecolman:
Move the storing of the flux from unsplit to cmpflxm. This avoids doing the additional slow store/load operations to/from the intermediate arrays fx and tx, which are no longer needed.


rteyssier (approved):
Seems like a good optimization. Tests worked so all good for me.

tinecolman:
I will just add the same for mhd, to keep things synced.

tinecolman:
One of the sink tests (stellar-HII) failed. I will investigate.

PR #148, opened 2025-06-22
Fix incorrect T2min calculation in cooling_fine.f90 when jeans_ncells > 0

cosmos-in-us:
This PR corrects the formula for calculating the temperature floor `T2min` in `cooling_fine.f90`, based on Truelove et al. (1998) criterion:

$$\lambda_J \ge N_J \Delta x_\text{lmax}$$,

where $\lambda_J$ is the Jeans length, $N_J$ is the minimum number of cells per Jeans length (`jeans_ncells` in the code), and $\Delta x_\text{lmax}$ is the cell size at the highest $\text{lmax}$ (`nlevelmax` in the code).

### Problem
The previous formula gave unphysical results when `jeans_ncells > 0`.
I encountered this issue specifically when using `jeans_ncells = 4`.
The problem stems from the following:

1. **Comoving vs. physical coordinates**: The cell size was computed using `scale_l/aexp`, but this PR changes it to the physical coordinates.
2. **Unit mismatch**: `nH` is already in cgs units between line 307 and line 539, so using `scale_d` in the `T2min` calculation was redundant.
3. **Missing constants**: The formula for `T2min` was missing the necessary constants (i.e.,`mH**2/X/kB` in line 340).

### Fix
I fixed both `polytropic_constant` and `T2min` in the code to properly satisfy the Truelove criterion. The updated formula ensures physically consistent behavior.

### Result
Gas phase diagrams before and after the fix are shown below:
T2min_before_Fix
T2min_after_Fix

### How to Validate
This test confirms that the corrected `T2min` formula sets the temperature floor high enough to meet the Truelove criterion at $log_{10} n_H = 0\ \mathrm{cm^{-3}}$, which corresponds to where the updated `T2min` curve intersects in the phase diagram.  

```py
# physical constant
G = 6.67e-8 # cgs
mH = 1.67373522e-24 # cgs
kB = 1.380649e-16 # cgs
X = 0.76
mu_H = 1/X
cm2kpc = 3.240730118347575e-22

gamma = 1.666666 # from namelist params
levelmin = 7 # from namelist params
nlevelmax = 14 # from namelist params

# set test paraemters
T = 10**(3.54) # K; temperature
nH = 1 # cm^{-3}; Hydrogen number density

rho = mu_H * mH * nH # g cm^{-3}; density
cs2 = gamma*kB*T*X/mH # cm/s; sound speed

d_jeans = np.sqrt(np.pi*cs2/G/rho) # cm; jeans length
print(f"For T = {T:.0f} K, and nH = 1 cm^3, ")
print(f"jeans length = {d_jeans*cm2kpc:.4f} kpc")

jeans_ncells = 4 # from namelist params
boxlen = 1
scale_l = 1.30345975e+25 # cm; from ds.length_unit
print(f"N_J * \Delta x = {boxlen * scale_l * 0.5**nlevelmax * cm2kpc * jeans_ncells:.4f} kpc")
```

**Output**:
```
For T = 3467 K, and nH = 1 cm^3, 
jeans length = 0.9021 kpc
N_J * \Delta x = 1.0313 kpc
```

This confirms that this PR is correct.

## PR Checklist

- [X] The pull request has a title and a description
- [X] For new code features, the PR should have been tested (please include the results of the tests below or, when possible, add a new test for the feature).


hydro/cooling_fine.f90
```diff
@@ -135,8 +135,7 @@ subroutine coolfine1(ind_grid,ngrid,ilevel)
 
   ! Polytropic constant for Jeans length related polytropic EOS
   if(jeans_ncells>0)then
-     polytropic_constant=2d0*(boxlen*jeans_ncells*0.5d0**dble(nlevelmax)*scale_l/aexp)**2/ &
-          & twopi*6.67d-8*scale_d*(scale_t/scale_l)**2
+     polytropic_constant=factG_in_cgs*(boxlen*jeans_ncells*0.5d0**dble(nlevelmax)*scale_l)**2/ pi / gamma
```
rteyssier: Your cell size is here in comoving coordinates. In cosmology, scale_l \propto aexp.
If you want to use a constant length scale physical coordinates, you need to use scale_l/aexp.

rteyssier: Otherwise your calculations are correct IMO.

cosmos-in-us: Thanks for your comments, Romain!

I'd like to clarify my understanding of `scale_l` and cell size. According to [units.f90](https://github.com/ramses-organisation/ramses/blob/56e4d182c042ab6680a5b6f2ba4811b17ec83191/amr/units.f90#L24), `scale_l` is in physical units and propotional to `aexp` (as you said). So, `scale_l/aexp` gives a comoving unit. Therefore, my cell size above is actually in physical coordinates.

I got your point about using the `scale_l/aexp` factor. The previous formula introduced it for keeping a constant physical length scale, not for converting it into comoving coordinates.

However, I'm still confused because, in my opinion, $\Delta x$ in Truelove condition needs to be a instantaneous physical cell size, so I thought, it naturally increases with `aexp`.

If the goal is to use a constant physical length scale for a specific purpose within the T2min calculation, then yes, dividing by `aexp` would indeed be the way to achieve that constant physical length scale. If the PR should aim for this constant physical, please let me know, and I can adjust it accordingly.

rteyssier:
On Ramses we have this trick that allows to unlock a new refinement level
when the expansion factors doubles. Tu is way we emulate a fixed physical
resolution even if we use comoving coordinates. The Jeans length formula is
based on the same constant physical coordinates resolution idea.

hydro/cooling_fine.f90
rteyssier: scale_l is the length scale in comoving units, not physical units. If you want a constant physical units, you need to divide by aexp.

cosmos-in-us:
The `polytropic_constant` formula was updated to ensure a constant physical cell size by dividing by `aexp`!

hydro/cooling_fine.f90
rteyssier: I think it is correct now. Note that one could additional parameters here like setting delta x equal to the comoving or physical cell size at levelmax (following the progressive unlocking of new levels with time), or alternatively to the current level comoving cell size. These extra parameters coud be added later though.

rteyssier (approved):

PR #151, opened 2025-06-23
Fix nlevelmax_sink not working as intended

tinecolman:
Patrick Hennebelle reported that sinks were not forming properly when using this. He traced it to a problem with clump_mass4. I looked further into it and found places where nlevelmax should be replaced by nlevelmax_sink.

I also fixed the issue were nlevelmax_sink could not be combined with sink_refine.

I updated the sink_spawning test namelist to catch this issue.
I verified there are no issues for sink feedback, by adding nlevelmax_sink to the tests that cover these.


nbrucy (changes requested):
It's not clear in the code what `nlevelmax_sink` should do and when it's needed.  Maybe it should be better documented in docs/Sinks.md ?

tinecolman:
> It's not clear in the code what `nlevelmax_sink` should do and when it's needed. Maybe it should be better documented in docs/Sinks.md ?

A bit unrelated to this bugfix, but added.

nbrucy (approved):

PR #152, opened 2025-06-23
add the possibility to use self-gravity and analytical gravitational …

simonb282:
I add the negative gravity types which support self-gravity and an analytical gravitational potential.
The new types are the negative counterpart of the old gravity_types.

For example, `gravity_type = -2` corresponds to self-gravity with a point mass gravitational potential, when `gravity_type = 2` corresponds only to a point mass gravitational potential.

It supports new gravity_types because it uses `ABS(gravity_type)` to check what type is requested.


nbrucy:
New try with only the needed changes  upcoming.

PR #153, opened 2025-06-23
More flexible gravity sources

simonb282:
This PR splits the `gravity_type` parameter in three individual parameters, allowing for more control over what sources of gravity can be added to the simulation:
* `self_gravity=.true.`: switch on self-gravity (previously `gravity_type<=0`)
* `gravity_rho_ana_type>0`: add an analytical contribution to the poisson source term using the routine `rho_ana` (previously `gravity_type<0` which automatically enabled self-gravity)
* `gravity_force_ana_type>0`: add an analytical contribution directly to the force (previously `gravity_type>0` and could not be combined with self-gravity)

To accommodate this change, the gravity_params namelist parameter has been replaced by `gravity_rho_ana_params` and `gravity_force_ana_params`.

This offers the choice to describe the gravity source with either the density distribution or the force, which is useful if one is easier to do that the other. It also allows combining the two types of analytical sources.
The motivation for this was to enable setups where there is an external tidal force outside the computational domain which affects the self-gravitating gas inside the box.

This change is backwards compatible. When the old parameters are found upon reading the namelist, the new parameters are set to recover the old behaviour.


poisson/gravana.f90
```diff
@@ -24,7 +24,7 @@ subroutine gravana(x,f,dx,ncell)
   real(dp)::scale_l,scale_t,scale_d,scale_v,scale_nH,scale_T2
 
   ! Constant vector
-  if(gravity_type==1)then
+  if(ABS(gravity_type)==1)then
```
nbrucy: Beware, this means that both rho_ana and gravana will be used ... Another switch mechanism should be used

nbrucy (changes requested):

rteyssier (changes requested):
Status of this pull request unclear...
Have the required changes been done?

nbrucy:
Hi Romain. Yes the change has been done. What is missing now is a bit of documentation, if merged as is I'm afraid the gravity selection will become more obscure.

tinecolman:
I've taken over this PR and made it more general.

nbrucy (approved):

rteyssier (approved):

PR #154, opened 2025-06-24
Refactor and optimize uslope

tinecolman:
I've gather the formulas for the different slope limiters into a module called slope_types. For each slope type, there is a pure function which the compiler should inline thanks to the additional compiler flags in the Makefile.
This structure reduces code duplication and makes it easier to optimize the existing slope limiters.

I've rewritten minmod/average for avoid the division and casting from int to float in the inner loop. This also makes the calculation of dlft and drgt the same as for the other regular slope types.

The slope module also contains a subroutine for each slope_type which is doing the calculation for nvector elements. Most of these call their corresponding slope_type function for each dimension.
In these subroutines, the center value is now prefetched into a scalar, and reused for the calculation of dlft and drgt, reducing the number of load operations. This gives a small performance gain.

The main subroutine uslope is now effectively a wrapper to call the requested calc_uslope subroutine.
I've rearranged the parts for different NDIM to have less duplication. For HYDRO slope_type 1, the structure is a bit strange due to their being a different formulation for 3D for unclear reasons. Since I didn't want to change the functionality, I've kept both.
 
For MHD, the calculation of the B-field slopes are now gathered at the end of the routine, to make it easier to merge hydro and mhd files in the future.


tinecolman:
An issue is that not all of the relevant slope_type - NDIM combinations are currently covered by the test suite.

nbrucy:
> An issue is that not all of the relevant slope_type - NDIM combinations are currently covered by the test suite.

So maybe we need to finish #141 before merging this one?

tinecolman:
I get a speedup of 5% on the uniform grid sedov benchmark with this.

tinecolman:
Something is not working for intel compiler. Currently investigating.

tinecolman:
For intel, the performance is the same as before: no gain but also no loss.

tinecolman:
I fixed a bug for slope_type=4, which was revealed by the sod parameter sweep.
Tests confirm that now everything works as before for the case HYDRO 1D.

tinecolman:
Using the orzag-tang test, I manually confirmed that the new implementation is now working as before for 2D MHD, with all slope_type options (1,2,3).

tinecolman:
Using the collapse-baro test, I manually confirmed that the new implementation is working as before for 3D MHD, with all slope_type options (1,2,3,7,8).

tinecolman:
Using the implosion, mixing, cooling-frig and sedov3d tests, I manually confirmed that the new implementation is now working as before for 2D and 3D HYDRO with all slope_type options (1,2,3,7,8).

tinecolman:
I've moved the if-check inwards, so that the outer loop i,j,k is now only done once. Then I created vectorized subroutines for each slope_type, so that uslope is now effectively a wrapper routine to select the requested slope type.
By keeping the dimensions of input and output arrays the same as before, the compiler properly inlines and vectorizes. And so the execution time is unaffected.

This structure will allow to move the call to uslope inside trace, bypassing a slow store-load operation. This change will be made in a separate PR once this one, and the refactoring of trace is accepted

rteyssier (approved):

rteyssier:
It looks like we have some conflicts here...

tinecolman:
> It looks like we have some conflicts here...

It's because I deleted mhd/hydro_parameters.90 in the other PR. Fixed it!

PR #157, opened 2025-06-24
Removed unused variables from godunov solver

tinecolman:
Remove gathering of  ploc, req_loc and peq_loc in godine1, as these are never used in unsplit. These might have come from an incomplete port of the mom2 and global_star patches.

### DIFF
--- a/hydro/godunov_fine.f90
+++ b/hydro/godunov_fine.f90
@@ -488,10 +488,7 @@ subroutine godfine1(ind_grid,ncache,ilevel)
   integer ,dimension(1:nvector,1:threetondim     ),save::nbors_father_cells
 
   real(dp),dimension(1:nvector,iu1:iu2,ju1:ju2,ku1:ku2,1:nvar),save::uloc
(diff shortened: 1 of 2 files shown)


rteyssier:
These variables are used in the global_star patch. Should not be removed.

rteyssier (changes requested):
This will prevent using the global_star patch.
I don't approve.

tinecolman:
I rebased the branch to include the recent changes in the hydro files.

tinecolman:
> This will prevent using the global_star patch. I don't approve.

I had a look at the global_star_patch. It is using the MHD solver. Here, I only update the HYDRO solver files. So in principle the global_star_patch should still work fine. That said, it's going to be a lot of work keeping this patch synchronized with the optimizations I'm making in the hydro solver. We should think about what to do with it. If it is important to have this patch working, we could move it to a branch?

rteyssier:
Another option that many users requested is to have a new set of variables called static_scalars that are used instead of plot, neq_loc etc...

tinecolman:
But the variables here don't behave like scalars?

rteyssier:
They are not passively advected like passive scalars. They are fixed. Or
static. Hence static scalars.

PR #158, opened 2025-06-24
Optimize ctoprim

tinecolman:
Remove the calculation of the sound speed from ctoprim since it is not used further in the muscle scheme. I moved it to the upldme scheme.

Second, remove store-load dependency by first retrieving the variables in a scalar and then writing them at the end after the energy computation.


tinecolman:
This gives me a 5.4% speedup on the sedov on marenostrum and 3% on meluxina.

rteyssier (approved):

PR #159, opened 2025-06-30
optimizations for hydro solver on uniform grid

tinecolman:
* Add the constants defining the loop bounds for the hydro stencil to the amr_constants module
* Move the gathering of the stencil from godfine1 to its own routine "gather_stencil_amr"
* Add a version of gather_stencil for uniform grid runs. 
 
For unigrid simulations, this circumvents the non-vectorizable indirect access caused by splitting up the cells into those that need interpolation and those that don't. We're still working on a solution for AMR runs.

This changes results in a 10% speedup for unigrid hydro simulations.
check_refactor


tinecolman:
Intel seems to have an issue with moving the gathering of the stencil to a separate subroutine. I'm looking further into it.

rteyssier (approved):

PR #168, opened 2025-07-10
fix Cosmo test fails to stop when testing with restart #167

### LINKED ISSUE #167: Cosmo test fails to stop when testing with restart
The end time is set to aout=1.355E-01. When executing the test with the restart mechanism (-r), it continuous to run after dumping output 2 at the specified aout. When running it the regular way, it stops after this output.

nbrucy:
Fixes issue #167.
Also better compares the clumps, discarding info on the index that is no guaranteed to be conserved after a restart.


tinecolman:
I confirm that this solves issue #167

tinecolman (approved):

tinecolman (approved):

PR #170, opened 2025-07-12
Refactor and document trace

tinecolman:
Full refactoring of the trace routines for the umuscl scheme in the godunov solver (hydro solver only).
* Combine trace1d, trace2d and trace3d into a single trace routine, removing code duplication.
* Move loop over nvar inside the vectorization loop, allowing for generalized code. These loops are unrolled as to behave similarly to the scalar instructions from before.
* Add loops over ndim, allowing to remove preprocessor directives. These loops are also unrolled.
* SIMD compiler directive added to ensure the vectorization is done on the correct loop. Despite this, the vectorization rate of the trace routine is reduced. Attempts at restoring the vectorization resulted in performance loss.
* Rewrite the calculations to reduce the number of math operations:
    * use only dtdx since dx=dy=dz.
    * precalculate /r once, then reuse.
* Add comments describing each term in the calculation.
* Add documentation at the top of the routine.

In the end, the performance is the same as the old version for the sedov benchmark, when measure on Meluxina with GNU.

WARNING: this PR is touching a performance heavy part of the hydro solver. If you experience any problems with compilation or performance related to this PR, please contact me.


tinecolman:
NENER + 2D/3D currently not covered by the test suite.

bin/Makefile
```diff
@@ -10,7 +10,7 @@ GCOV = 0
 # Compiler flavor: GNU or INTEL
 COMPILER = GNU
 # Size of vector cache
-NVECTOR = 32
+NVECTOR = 16
```
rteyssier: Do we want to change this?

bin/Makefile
```diff
@@ -108,7 +108,7 @@ ifeq ($(COMPILER),GNU)
          F90 += -g -O0 -fbacktrace -fbounds-check -Wuninitialized -Wall
          FFLAGS += -ffpe-trap=zero,underflow,overflow,invalid -finit-real=nan
       else
-         F90 += -O3
+         F90 += -O3 -march=native -flto -fwhole-program -funroll-loops -g
```
rteyssier: These options seem specific. Do we want to add them to the default makefile?

bin/Makefile
tinecolman: I think we do. I've been finding consistently better speeds with 16. But it should probably not be in this PR.

bin/Makefile
tinecolman: Some yes, some no. In any case, it should not go in the PR. Except maybe the unroll-loops one, which may have been needed for this rewrite. I should check.

jokirosdahl (approved):

rteyssier (approved):

PR #172, opened 2025-07-16
update dice patch

tinecolman:
Synchronised the dice patch with the current dev branch. Also made init_flow_fine.f90 work with MHD (hopefully...).


tinecolman:
It compiles with both hydro and mhd. But I have no test to see if it is actually working or not.

mtrebitsch:
It works well with hydro: I can reproduce the [idealised galaxy tutorial](https://ramses-tutorials.readthedocs.io/en/latest/Idealised-Disc-Galaxy/tuto_disc.html), I will try to test some MHD case ASAP.

mtrebitsch (approved):
Tested with the idealised galaxy test from the [RAMSES tutorials](https://ramses-tutorials.readthedocs.io/en/latest/Idealised-Disc-Galaxy/tuto_disc.html), and with a constant magnetic field from the DICE patch itself.

In both cases, the code runs. In the hydro case, the results are comparable to a similar setup with the old version. I had no comparable test for the MHD case (and little experience with MHD runs), but the code behaves "as expected".

nbrucy (approved):

PR #173, opened 2025-07-17
Star formation, SN feedback and chemical enrichement test

pocvirk:
This contribution adds a test for star formation, SN kinetic feedback and chemical enrichement.
It adds a dir star-formation-and-feedback in the tests dir
It is based on the poisson/cosmo setup but:
It compiles with NMETALS=1 
It  adds params for star formation, SN feedback and chemical enrichment in the .nml
It tests the output by extracting a subcube around the highest density region of the box (where the SN FB happens), zooming in, plots density, temperature and absolute metallicity, and comparing the stats of that cube against the ref results.


rteyssier (changes requested):
A lot of weird files...

tinecolman:
This PR became a bit of a mess. This will be fixed once the final version of the test is complete.

PR #177, opened 2025-07-30
Correction of bugs in condinit_kind='collapse'

matthiasglz:
Corrects bugs in condinit_kind='collapse' (scaling factors wrong for omega0 and B0)

Adds the possibility to change directly in the namelist :
- the core mass (mass_c)
- the angle between the magnetic field and the rotation axis (theta_mag)
- an initial turbulent velocity field (Mach)

Updates accordingly the collapse-baro reference solution in the test suite

Capture d’écran 2025-07-30 à 17 26 13

## PR Checklist

- [X] The pull request has a title and a description
- [X] For new code features, the PR should have been tested (please include the results of the tests below or, when possible, add a new test for the feature).
- [X] The code covered by the pull request is documented


mhd/condinit.f90
```diff
@@ -187,24 +200,112 @@ subroutine collapse_condinit(x,q,dx,nn)
...
+    vy_tot=0.d0
+    vz_tot=0.d0
+    vx2_tot=0.d0
+    vy2_tot=0.d0
+    vz2_tot=0.d0
+    v_rms=0.d0
+    count_vrms=0
+    if(Mach .ne. 0)then
+      if (myid==1) write(*,*) 'Read the file which contains the initial turbulent velocity field'
+      open(20,file='init_turb.data',form='formatted')
```
nbrucy, line 255: I think this file is missing (also maybe the name of the file could be a parameter)

nbrucy (dismissed):

mhd/condinit.f90
matthiasglz: Yes, I didn't include this file because it is specific to certain initial conditions, and I thought it was up to each person to choose the initialisation file they wanted.
The test suite collapse doesn't need it.
It was just so that the functionality would be present in the condinit file.

rteyssier (approved):

PR #185, opened 2025-08-28
Overhaul of compiler flags in Makefile

tinecolman:
Separate compiler flags in the Makefile in 3 categories:
* FFLAGS_BASE: base compilation flags to make the code run in a specified mode
* FFLAGS_OPT: optimization compilation flags
* LDFLAGS: flags that need to be applied for both compilation and linking
F90 now contains only the pure compiler command, e.g. gfortran, ifort or their MPI versions.

Added a supplementary Makefile called clusters.mk for adding machine specific compiler flags. Moved the MPI compiler command MPIF90 to this file. Filled in the value of MPIF90 for 7/8 EuroHPC clusters. Optimization flags to be added later.

The parameters in clusters.mk will be loaded automatically by the Makefile when setting the new parameter MACHINE at the top. DEBUG and COV options will overwrite optimisation flags set in clusters.mk.

### DIFF
--- a/bin/Makefile
+++ b/bin/Makefile
@@ -86,52 +86,72 @@ ifeq ($(LIGHT_MPI_COMM), 1)
    DEFINES += -DLIGHT_MPI_COMM
 endif
 #############################################################################
-# Fortran compiler options and directives
+# Compiler command for regular compilation (mpi or serial)
+F90 =
+# Serial compiler command and flags for helper files
+FC =
+# Base compiler flags for compilation
+FFLAGS_BASE =
+# Compiler flags for optimization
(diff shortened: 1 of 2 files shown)


rteyssier (approved):

jokirosdahl:
My Macbook M1 is not happy with the new -flto option in the makefile

tinecolman:
> My Macbook M1 is not happy with the new -flto option in the makefile

So I've heard but I will need more information to be able to fix the problem.

rteyssier:
Same thing here for my M1 older MacBook.
I think this option should be optional not default.

PR #188, opened 2025-09-09
remove flto as a default compiler flag to fix compilation on mac

tinecolman:
The flto optimization flags are moved to the presets of clusters.
This should fix the failed compilation on arm64 processors, which can be found in macbooks.


tinecolman:
This PR provides a fix for issue #189 until a better solution is found.

rteyssier (approved):

tinecolman:
The issue #189 was fixed by PR #190

PR #196, opened 2025-09-23
Openmp virtual_boundaries

tinecolman:
Add OpenMP to the communication routines in virtual_boundaries.f90.
We choose to put the openmp directive on the `do j=1,twotondim` loops, which have been moved outwards. This way, each threads operates only on their own data.


sanhancluster (commented):
I guess cache arrays (e.g., ind_cell, ind_grid, xx) need to be also added to private variables for the parallel region?

tinecolman:
> I guess cache arrays (e.g., ind_cell, ind_grid, xx) need to be also added to private variables for the parallel region?

Indeed! They have already been added in a previous PR.

PR #197, opened 2025-09-23
[Tests] add cosmo test with amr

tinecolman:
This test is an alternative version of the cosmo test, with 1 amr level. It helps in verifying the OpenMP implementation.


tinecolman:
No longer needed, since PR#198 covers cosmo with amr

PR #198, opened 2025-09-29
added star formation and SN feedback test

pocvirk:
This PR uses the cosmo test, runs it with AMR until stars form (model 0), and then a couple of SN explode (kinetic model).
A new namelist parameter is added to prevent random Poisson noise to be added to the star. This was needed to be able to reproduce the results on different machines.


tinecolman:
We will need to add the new star category to the test workflow for it to run on github actions.

pocvirk:
to deal with the rather large errors in this test:
I added a parameter randomize_sf=.true. by default but can be set to .false. for tests where randomization is to be avoided. Impacts star_formation.f90, read_hydro_parameters and amr_parameters. Also, added a star/sfonly dir where we just test star formation, not SN FB. Then SNFB is tested in its own dir, also with randomize_sf=.false."

pocvirk:
removed tests/star/sfonly from tracking and repo.

tinecolman:
Screenshot from 2025-10-02 14-53-10

This is what it looks like.

tinecolman (approved):

PR #202, opened 2025-10-03
Start arrays flag1 and flag2 from 1 instead of 0

tinecolman:
I couldn't find a reason why they should start from 0.

### DIFF
--- a/amr/flag_utils.f90
+++ b/amr/flag_utils.f90
@@ -31,7 +31,7 @@ subroutine flag_coarse
   ! Constants
   nxny=nx*ny
   ! Reset flag1 array at coarse level
-  flag1(0:ncoarse)=0
+  flag1(1:ncoarse)=0
   ! Set flag1 to 1 at coarse level for inner cells only
   nflag=0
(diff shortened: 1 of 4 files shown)


rteyssier:
This is because in one routine we use index 0 for non existing neighbors.

tinecolman:
> This is because in one routine we use index 0 for non existing neighbors.

Which routine? I could not find any place where index 0 is accessed but maybe I missed it.

rteyssier:
I think smooth_fine.

tinecolman:
> I think smooth_fine.

Ok, I see there is a call to count_nbors where we do
```
  ! Gather neighbors
  call getnborcells(igridn,ind,indn,nn)
  ! Check if neighboring cell exists and count it as a flagged neighbor
  i_nbor(1:nn)=0
  do in=1,twondim
     do i=1,nn
        i_nbor(i)=i_nbor(i)+flag1(indn(i,in))
     end do
  end do
```
where indn may indeed contain 0.

rteyssier:
Yep

PR #203, opened 2025-10-07
Combine the hydro and mhd version of init_hydro.f90

tinecolman:
Both versions were first synchronized. The mhd version was missing the passive scalar / nener remapping on restart and the strict equilibrium variables reading block.
Preprocessor directives have been added to the hydro version to account for MHD. The variable nvar2 is properly adjusted to take into account the presence of the right B-field at indices nvar+1 to nvar+3.


tinecolman:
My motivation for this is that I may want to change some things in the array initialisation for openMP.

tinecolman:
I've run the test suite with restart mechanism on this and confirmed the behavior is the same as before (pass or failed with the same errors).

rteyssier (approved):

PR #204, opened 2025-10-14
bugfix in quick_sort_int_int

tinecolman:
Fix a bug where order is uninitialized in the integer version of quick_sort. This causes the returned order book-keeping array to be incorrect (possibly all 0).
This routine is used in the merger_tree module.

### DIFF
--- a/amr/sort.f90
+++ b/amr/sort.f90
@@ -319,6 +319,13 @@ SUBROUTINE quick_sort_int_int(list, order, n)
   INTEGER, DIMENSION (1:n), INTENT(INOUT)  :: order
 
 
+  ! Local variable
+  INTEGER :: i
+
+  DO i = 1, n
+     order(i) = i
+  END DO
+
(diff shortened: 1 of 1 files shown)


rteyssier:
No this is not a bug. The order must be initialized by the calling routine.
If you make this change, this will break the merger tree code.

amr/sort.f90
```diff
@@ -319,6 +319,13 @@ SUBROUTINE quick_sort_int_int(list, order, n)
   INTEGER, DIMENSION (1:n), INTENT(INOUT)  :: order
 
 
+  ! Local variable
+  INTEGER :: i
+
+  DO i = 1, n
```
rteyssier, line 325: Must be done before in the calling routine.
The order does not have to be 1:n.

rteyssier (changes requested):

tinecolman:
I see. How silly of me to think the usage would be the same as quikc_sort_dp.

PR #206, opened 2025-10-17
Revert "Refactor and document trace"

tinecolman:
Reverts ramses-organisation/ramses#170

Some of the added preprocessor directives are causing problems, specifically with some versions of mpiifort.


tinecolman:
I'm bypassing the review, since this is a revert.

PR #218, opened 2025-12-01
Document the lightcone file

raserayann:
Add comments in the lightcone file


nbrucy:
Thanks! Maybe we should merge the bugfix directly into dev.

nbrucy (approved):

PR #223, opened 2025-12-10
[DO NOT MERGE] This PR is there to test the CI.

nbrucy:
To test the CI


tinecolman:
You should first check compilation, and only when everything compiles, you run the test suite. Why is build with RT in 3D etc taking more than 9 minutes?

nbrucy:
> You should first check compilation, and only when everything compiles, you run the test suite. Why is build with RT in 3D etc taking more than 9 minutes?

For the first point, that should be a next PR.
Downloading the dependencies strangely takes a very long time, maybe github servers are overloaded.

PR #232, opened 2026-01-12
Python to RAMSES input through grafic files + test

tinecolman:
This PR adds a python script to utils to write grafic-style input files. This allow to easily convert numpy arrays to binary input files that can be read with RAMSES. Currently this only generates input for hydro variables (non-cosmo runs).

An example of how this can be used is added in the form of a decaying isothermal turbulence test case. The initial conditions are grafic files, containing:
* a random velocity field
* a uniform density and pressure, such that T=10 K


tinecolman:
The CI will currently fail because turbustat is not installed. But that needs an older version of numpy.  So I wasn't sure how to deal with this.

nbrucy:
We can force the version of numpy in the requirements.

tinecolman:
Python packages problems for github actions fixed

utils/py/write_grafic.py
```diff
@@ -0,0 +1,82 @@
...
+    with open(filename, "ab") as f:  # append binary
+        for i in range(0,ncells):
+            # Ensure float32
+            a = np.asarray(data_temp[i,:,:], dtype=np.float32, order="F")
+
+            # convert byte order if needed
+            if ((a.dtype.byteorder == "=" and swap) or
+                (a.dtype.byteorder == ">" and endian == "<") or
+                (a.dtype.byteorder == "<" and endian == ">") ):
+                a = a.byteswap().newbyteorder()
```
cphyc: This could be a one-liner:

```suggestion
            # Convert input array into float32 with selected endianness
            dtype = np.dtype(f"{endian}f")
            a = data_temp[i,:,:].astype(dtype, order="F")
```

utils/py/write_grafic.py
```diff
@@ -0,0 +1,82 @@
...
+        n1, n2, n3,
+        dx, xoff1, xoff2, xoff3,
+        boxlen, f1, f2, f3)
+
+    reclen = len(payload)
+
+    with open(filename, "wb") as f:
+        f.write(struct.pack(endian+"i", reclen))  # leading record marker
+        f.write(payload)
+        f.write(struct.pack(endian+"i", reclen))  # trailing record marker
```
cphyc: Since we already depend on scipy, we can do `from scipy.io import FortranFile` and replace this with:

```suggestion
    with FortranFile(filename, "w") as f:
        f.write_record(n1, n2, n3, dx, xoff1, xoff2, xoff3, boxlen, f1, f2, f3)
```

utils/py/write_grafic.py
```diff
@@ -0,0 +1,82 @@
...
+    ''' Write data in slices to grafic binary file '''
+    ncells = data.shape[0]
+    data_temp=data.transpose(2,0,1)
+
+    # check if the native byte order matches the requested one
+    native_byteorder = sys.byteorder
+    swap = ( (endian == '<' and native_byteorder == 'big') or
+             (endian == '>' and native_byteorder == 'little') )
+
+    with open(filename, "ab") as f:  # append binary
```
cphyc: If we're using `FortranFile` from scipy, this would become:

```suggestion
    # Append in binary mode
    with open(filename, "ab") as file_desc, FortranFile(file_desc) as f:
```

utils/py/write_grafic.py
```diff
@@ -0,0 +1,82 @@
...
+import sys
+
+def write_grafic_header(filename, ncells, size, endian='<'):
+    ''' Write a simple header for the grafic binary file '''
+    # variables for the header line
+    n1 = n2 = n3 = int(ncells)
+    dx = float(size/ncells)
+    xoff1 = xoff2 = xoff3 = float(0.0)
+    boxlen = float(size)
+    f1 = f2 = f3 = float(0.0)
```
cphyc: This is misleading since `int` in Python is actually an int64 and `float` is a float64. I'd suggest

```suggestion
    n1 = n2 = n3 = np.int32(ncells)
    dx = np.float64(size/ncells)
    xoff1 = xoff2 = xoff3 = np.float64(0.0)
    boxlen = np.float64(size)
    f1 = f2 = f3 = np.float64(0.0)
```

utils/py/write_grafic.py
```diff
@@ -0,0 +1,82 @@
...
+                (a.dtype.byteorder == ">" and endian == "<") or
+                (a.dtype.byteorder == "<" and endian == ">") ):
+                a = a.byteswap().newbyteorder()
+
+            raw = a.tobytes(order="F")
+            reclen = len(raw)
+
+            f.write(struct.pack(endian + "i", reclen))
+            f.write(raw)
+            f.write(struct.pack(endian + "i", reclen))
```
cphyc: With FortranFile this then would become:

```suggestion
            f.write_record(a)
```

utils/py/write_grafic.py
```diff
@@ -0,0 +1,82 @@
+'''
+    Module to write grafic binary files used as initial conditions for
+    non-cosmo simulations with RAMSES.
+
+    Be careful in cases where the order of the array representation matters!
+    Little/big endian issues are NOT dealt with!
```
cphyc: Aren't they?

tests/hydro/decaying-turbulence/initial_conditions.py
```diff
@@ -0,0 +1,109 @@
...
+sound_speed = (kb_cgs*T/(mu*mH_cgs))**0.5 / scale_v # code units
+
+''' ---- Generate ICs ---- '''
+
+# generate fields for the three components of the velocity
+# the average velocity will be zero (no bulk motion)
+# the velocity dispersion will be 1 (before rescaling to Mach)
+vx = make_3dfield(ncells, powerlaw=powerlaw, randomseed=seed1)
+vy = make_3dfield(ncells, powerlaw=powerlaw, randomseed=seed2)
+vz = make_3dfield(ncells, powerlaw=powerlaw, randomseed=seed3)
```
cphyc, line 107: I think this is the only spot where we use turbustat. Is it worth it given that we need to pin a low version of numpy (soon to be deprecated)?

Otherwise, we could implement `make_3dfield` with
```python
def make_3dfield(ncells: int, *, powerlaw: int, randomseed: int):
    """Generate a 3D random field with a power-law power spectrum.

    Parameters
    ----------
    ncells : int
        The number of cells along each dimension of the 3D field.
    powerlaw : int
        The exponent of the power-law power spectrum. For example, a value of
        2 would correspond to a power spectrum that scales as k^-2.
    randomseed : int
        The seed for the random number generator, ensuring reproducibility of
        the generated field.

    Returns
    -------
    field : numpy.ndarray
        A 3D array of shape (ncells, ncells, ncells) containing the generated
        random field values.
    """
    shape = (ncells, ncells, ncells)
    shape_fourier = (ncells, ncells, ncells // 2 + 1)

    # Generate the Fourier space whitenoise
    generator = np.random.default_rng(randomseed)
    args = {"scale": 1 / np.sqrt(2), "size": shape_fourier}
    whitenoise = generator.normal(**args) + 1j * generator.normal(**args)

    # Create the k-space grid
    kx = np.fft.fftfreq(ncells) * ncells
    ky = np.fft.fftfreq(ncells) * ncells
    kz = np.fft.rfftfreq(ncells) * ncells
    kx, ky, kz = np.meshgrid(kx, ky, kz, indexing="ij")
    k = np.sqrt(kx**2 + ky**2 + kz**2)

    # Compute power spectrum
    Pk = np.zeros_like(k)
    Pk[k > 0] = k[k > 0] ** -powerlaw

    # Apply the power spectrum to the whitenoise
    field_fourier = whitenoise * np.sqrt(Pk)

    return np.fft.irfftn(field_fourier, s=shape)
```

utils/py/write_grafic.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Used for the decaying turbulence test case in tests/hydro/decaying-turbulence
+'''
+
```
tinecolman:
```suggestion
from scipy.io import FortranFile
```

utils/py/write_grafic.py
tinecolman: > Aren't they?
That's a remnant of before it was dealt with:)


```suggestion
```

tests/hydro/decaying-turbulence/initial_conditions.py
tinecolman: This seems like a good alternative. I'll give it a try, thanks!

matthiasglz: The dense core collapse tutorial also uses turbustat. If this alternative works, it would be worth updating the tutorial to remove the dependency.

nbrucy (changes requested):
I agree that it's better to avoid to pinpoint an old numpy version if there is an alternative.

utils/py/write_grafic.py
tinecolman: Should be float32 actually

tests/hydro/decaying-turbulence/initial_conditions.py
tinecolman: I've added the routine. Seems to be working well! I just had to add the additional correction since this doesn't seem to guarantee that vbulk=0 and the dispersion is not 1.

tests/hydro/decaying-turbulence/initial_conditions.py
```diff
@@ -0,0 +1,159 @@
...
+# make plots to verify the fields look OK
+for seed, v in zip([seed1, seed2, seed3], [vx,vy,vz]):
+    fig, ax = plt.subplots(nrows=1, ncols=3, figsize=[10, 3])
+    ax[0].imshow(v.mean(0), origin='lower')
+    ax[1].imshow(v.mean(1), origin='lower')
+    ax[2].imshow(v.mean(2), origin='lower')
+    plt.tight_layout()
+    plt.savefig('seed{}_pl{}.png'.format(seed, powerlaw))
+    plt.close()
+'''
```
nbrucy:
```suggestion

```
remove?

nbrucy (changes requested):

tests/hydro/decaying-turbulence/initial_conditions.py
tinecolman: No, it needs to stay in case you want to check the full initial velocity field

nbrucy (approved):

tests/hydro/decaying-turbulence/initial_conditions.py
cphyc: Then maybe add it as an option instead of a comment :) ?

tinecolman: As you wish

tinecolman:
Thanks @cphyc for the input on the grafic python script. This allowed to simplify the thing even more!

cphyc (approved):

PR #242, opened 2026-02-12
Fix galic util and update merger patch

tinecolman:
Add missing extern keyword that causes compilation error for galic.
Update files in merger patch for constants module and condinit_kind.

## PR Checklist

- [ ] The pull request has a title and a description
- [ ] For new code features, the PR should have been tested (please include the results of the tests below or, when possible, add a new test for the feature).


patch/hydro/merger/units.f90
```diff
@@ -1,33 +1,31 @@
...
   !-----------------------------------------------------------------------
 
   ! scale_d converts mass density from user units into g/cc
   !scale_d = 1.0
-  masse_solaire = 1.9891D33 ! g
-  kparsec = 3.085677581282D21 ! cm
-  ! On prend comme unit� de masse : 10**9 masses solaires
-  scale_d = 1.0D9 * masse_solaire / kparsec**3
-  if(cosmo) scale_d = omega_m * rhoc *(h0/100.)**2 / aexp**3
+  ! On prend comme unit de masse : 10**9 masses solaires
```
nbrucy:
```suggestion
  ! Unit of mass : 10**9 solar masses
```

utils/ic/galaxy_ic/Galic/globvars.h
```diff
@@ -114,7 +114,7 @@ extern double    *vxp_gas,*vyp_gas,*vzp_gas;
 
 
 
-double  LL;       /* LL = extension of fields in R and z. */
+extern double  LL;       /* LL = extension of fields in R and z. */
```
nbrucy, line 117: Why is this `extern` keyword needed?

nbrucy (changes requested):

utils/ic/galaxy_ic/Galic/globvars.h
tinecolman: Because otherwise the thing doesn't compile. If you look into the other files, you can see they all have the extern keyword. It's because they are defined elsewhere and then used in various files. It's ugly but nothing I can do about without rewriting the thing.

nbrucy (approved):

PR #245, opened 2026-02-18
First sfvirial=.true. test

pocvirk:
This test draws from the idealised galactic disc tutorial.
- It DLs DICE ICs from PO's seafile
- it runs the test
- compares with reference data


pocvirk:
mistake, wrong branch...

PR #249, opened 2026-02-23
changed the default load-in in amr/init_time.f90 from deltab to velcx…

James11222:
I changed the default load-in in `amr/init_time.f90` from `ic_deltab` to `ic_velcx` so that `monofonIC` ICs with no baryons are usable by ramses. Currently when `DoBaryons = False` in `monofonIC`, there is no `ic_deltab` file which means ramses throws an error and can't do the dmo sim, this fixes that issue since `ic_velcx` is present in dmo and dm+baryon cosmological simulation ICs.

## PR Checklist

- [x] The pull request has a title and a description
- [x] The code covered by the pull request is documented

### DIFF
--- a/amr/init_time.f90
+++ b/amr/init_time.f90
@@ -450,9 +450,9 @@ subroutine init_cosmo
         if(initfile(ilevel).ne.' ')then
            if(multiple)then
               call title(myid,nchar)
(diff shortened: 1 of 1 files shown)


tinecolman (approved):

tinecolman:
Seems like a good reason to change it. Hopefully it doesn't break other people's setup. Maybe we should think about making IC reading more general at some point.

PR #253, opened 2026-02-26
More flexible nener options

tinecolman:
* Add an option `io_nener_energies` to output/restart nener fields as energies, even if write_conservative=.false.
* Add the option to specify nener energy in the initial conditions region parameters using `Erad_region` instead of `prad_region`. Switching between the two is controlled by `io_nener_energies.
* Add the option to set boundary conditions for nener variables, using `prad_bound` or `Erad_bound`. The variable `prad_bound` was previously defined but not used.
* Add possibility to refine on gradients in the sum of all nener fields with `err_grad_prad` and `err_grad_Erad`. Also added the accompanying  `floor_prad` and  `floor_Erad`.


rteyssier (approved):

tinecolman:
I'm holding off merging this because it may make the FLD merge more difficult instead of easier...

PR #262, opened 2026-04-15
 Fix non-convergence in the H2 fraction for zero-metallicity gas

jokirosdahl:
Stabilizing fixes for the computation of equilibrium ionisation fractions in the non-equilibrium chemistry module.
This fixes non-convergence in the H2 fraction for zero-metallicity gas.

Adds a test for the routine cmp_Equilibrium_Abundances


tests/rt/eq-abundances/eq-abundances.nml
```diff
@@ -0,0 +1,57 @@
...
+y_center=0.25, 0.25, 0.75, 0.75
+length_x=0.5, 0.5, 0.5, 0.5
+length_y=0.5, 0.5, 0.5, 0.5
+exp_region=10
+d_region=3d-2, 3d0, 3d-2, 3d0      ! cm-3
+p_region=  0.782814, 78.282,  2609.38, 260938.5    ! 3e3, 3e3, 1e7, 1e7 K
+/
+
+&OUTPUT_PARAMS
+! 100 outputs, evenly log-spaced, ending at 500 myrs
```
tinecolman: This comment can be removed I think?

tests/rt/eq-abundances/plot-eq-abundances.py
```diff
@@ -0,0 +1,40 @@
...
+plt.tight_layout(pad=0.1)
+plt.savefig("eq-abundances.pdf", bbox_inches='tight')
+
+data = {"data": {}}
+
+data["data"]['T2']     = T2
+data["data"]['xHI']    = xHI
+data["data"]['xHII']   = xHII
+data["data"]['xHeII']  = xHeII
+data["data"]['xHeIII'] = xHeIII
```
tinecolman, line 40: Can you add also xH2 and xHeI to be checked?

rt/rt_init_xion.f90
```diff
@@ -232,24 +232,70 @@ SUBROUTINE cmp_Equilibrium_Abundances(T2,nH,phI_rates,mu,nSpec,Zsolar)
...
      else
-        mu_left =mu_left
-        mu_right=0.5*(mu_left+mu_right)
+        mu_right= mu_mid
      end if
-     err_mu=ABS(err_mu)
+     err_mu = ABS(mu_right-mu_left)/mu_mid
      niter=niter+1
   end do
+  mu=mu_mid
```
tinecolman: mu wasn't updated at the end of this routine before?

rt/rt_init_xion.f90
```diff
@@ -232,24 +232,70 @@ SUBROUTINE cmp_Equilibrium_Abundances(T2,nH,phI_rates,mu,nSpec,Zsolar)
...
+  integer::i
+
+! Test cmp_Equilibrium_Abundances routine and write the output to a file
+! Run for one gas density with no radiation field and zero metallicity
+! (this tends to be the setup most prone to non-convergence fot the
+! iterative calculation, due to the weak H2 formation rate at low T)
+!-------------------------------------------------------------------------
+  T20 = 1d0
+  T21 = 6d0
+  nH=1e0 ! 1 atom per cm3
```
tinecolman, line 276: Why use a hard-coded value instead of the actual value in the grid set by d_region in the namelist?

tinecolman (changes requested):

rt/rt_init_xion.f90
jokirosdahl: Good spot. It is updated through the call to cmp_chem_eq, but it means that this assignment is unnecessary and actually strictly wrong (though it has no consequences). I will remove it.

rt/rt_init_xion.f90
jokirosdahl: I'm actually not using d_region at all in this test. I'll just remove it from the namelist.

tests/rt/eq-abundances/plot-eq-abundances.py
jokirosdahl: Will do!

tests/rt/eq-abundances/eq-abundances.nml
jokirosdahl: Done

tinecolman (approved):

rteyssier (approved):

PR #276, opened 2026-06-26
small fix in MPI for sinks

tinecolman:
Correct root for MPI_BCAST for sink properties. As it, there would be a problem if you run with only 1 MPI-rank.

Bug found by @matthiasglz


nbrucy (approved):
I checked with a regex that was actually the only place where MPI_BCAST was call with a nonzero root.

rteyssier (commented):
OK for me.

PR #277, opened 2026-06-26
fix bug in time interpolation of turbulence force field

tinecolman:
The impact of this bug is relatively minor, though it depends on wheter dt_turb is smaller or larger than dt_hydro, and on whether the autocorrelation time is large enough compared to dt_turb.

I've updated the test solutions which use the TURB module.

Bug found by @rteyssier


turb/turb_check_time.f90
```diff
@@ -4,7 +4,6 @@ subroutine turb_check_time
    implicit none
 
    real(kind=dp) :: turb_last_tfrac        ! Time fraction since last
```
nbrucy:
```suggestion
   real(kind=dp) :: turb_tfrac        ! Time fraction between last and new turbulence field evaluation
```

turb/turb_check_time.f90
```diff
@@ -32,9 +31,9 @@ subroutine turb_check_time
       end do
 
       turb_last_tfrac = real((t - turb_last_time) / turb_dt, dp)
```
nbrucy:
```suggestion
      turb_tfrac = real((t - turb_last_time) / turb_dt, dp)
```

turb/turb_check_time.f90
```diff
@@ -32,9 +31,9 @@ subroutine turb_check_time
...
-      turb_next_tfrac = 1.0_dp - turb_last_tfrac
 
-      afield_now = turb_last_tfrac*afield_last + turb_next_tfrac*afield_next
+      ! interpolate for current time between last and next turb field
+      afield_now = (1.0_dp - turb_last_tfrac)*afield_last + turb_last_tfrac*afield_next
```
nbrucy:
```suggestion
      afield_now = (1.0_dp - turb_tfrac)*afield_last + turb_tfrac*afield_next
```

nbrucy (changes requested):
The test shows that the error was about 1 %.
I just suggest renaming the variable to avoid confusion.

tinecolman:
For the test setup the error is 1%, but not sure how big the error would be for different parameters.

nbrucy (approved):

PR #278, opened 2026-06-29
Add the option to output the electric current

tinecolman:
Reading and writing is managed by the new namelist parameters output_current and input_current.
The current is calculated in godfine1 from the magnetic field, and stored in the global cell-based array "electric_current".
When outputted, it is added as the last 3 variable.

### DIFF
--- a/amr/read_params.f90
+++ b/amr/read_params.f90
@@ -471,7 +471,8 @@ subroutine read_output_params(namelist_unit,nml_ok)
    ! Parameters to specify when to write an output
    namelist/output_params/noutput,foutput,aout,tout &
    & ,tend,delta_tout,aend,delta_aout,gadget_output,walltime_hrs,minutes_dump &
-   & ,output_to_log,write_conservative,read_conservative,exact_output_time
+   & ,output_to_log,write_conservative,read_conservative,exact_output_time &
+   & ,output_current,input_current
 
    ! Go to the beginning of the file
    rewind(namelist_unit)
--- a/mhd/godunov_fine.f90
+++ b/mhd/godunov_fine.f90
@@ -955,8 +955,6 @@ subroutine godfine1(ind_grid,ncache,ilevel)
   !-------------------------------------------------------
 #ifdef SOLVERmhd
   if(output_current) then
-     ! J = curl(B) at cell centres, using face-centred B from uloc
-     ! uloc(:,:,:,:,6)=Bx at left x-face, 7=By at left y-face, 8=Bz at left z-face
      do k2=k2min,k2max
      do j2=j2min,j2max
      do i2=i2min,i2max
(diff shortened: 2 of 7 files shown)


rteyssier (approved):
Looks good.

amr/read_params.f90
```diff
@@ -471,7 +471,8 @@ subroutine read_output_params(namelist_unit,nml_ok)
    ! Parameters to specify when to write an output
    namelist/output_params/noutput,foutput,aout,tout &
    & ,tend,delta_tout,aend,delta_aout,gadget_output,walltime_hrs,minutes_dump &
-   & ,output_to_log,write_conservative,read_conservative,exact_output_time
+   & ,output_to_log,write_conservative,read_conservative,exact_output_time &
+   & ,output_current,input_current
```
nbrucy, line 475: Could `input_current` be determined from the hydro_descriptor file instead of the namelist?

tinecolman: That requires parsing the descriptor, which is a much bigger change.

tinecolman:
I forgot to add the new params to the doc

mhd/godunov_fine.f90
```diff
@@ -950,6 +950,38 @@ subroutine godfine1(ind_grid,ncache,ilevel)
...
+     do i2=i2min,i2max
+        ind_son=1+i2+2*j2+4*k2
+        iskip=ncoarse+(ind_son-1)*ngridmax
+        do i=1,ncache
+           ind_cell(i)=iskip+ind_grid(i)
+        end do
+        i3=1+i2
+        j3=1+j2
+        k3=1+k2
+        do i=1,ncache
```
bcommerc, line 969: The curl is computed at different locations (on the left faces) and not at the cell center. 

something like 

electric_current(ind_cell(i),1) = (0.5*(uloc(i,i3,j3+1,k3,8)+uloc(i,i3,j3+1,k3,nvar+3)) - 0.5*(uloc(i,i3,j3-1,k3,8)+uloc(i,i3,j3-1,k3,nvar+3))/(2*dx) &
                                           -  (0.5*(uloc(i,i3,j3,k3+1,7)+uloc(i,i3,j3,k3+1,nvar+2)) - 0.5*(uloc(i,i3,j3,k3-1,7)+uloc(i,i3,j3,k3-1,nvar+2))/(2*dx)[10:36 AM]or compute dyBz=(0.5*(uloc(i,i3,j3+1,k3,8)+uloc(i,i3,j3+1,k3,nvar+3)) - 0.5*(uloc(i,i3,j3-1,k3,8)+uloc(i,i3,j3-1,k3,nvar+3))/(2*dx)
dzBy = (0.5*(uloc(i,i3,j3,k3+1,7)+uloc(i,i3,j3,k3+1,nvar+2)) - 0.5*(uloc(i,i3,j3,k3-1,7)+uloc(i,i3,j3,k3-1,nvar+2))/(2*dx)
etc….

should work. But I am not sure indexes +/- 1 always exist.  

The Bfield is first computed at the cell center of the neighbors, and then the curl is done over two cells. It is possible to do it over the cell, but the 3 components Bfield has to be computed on each face, which is a bit trickier, but doable.

bcommerc (changes requested):
The curl is computed at different locations (on the left faces) and not at the cell center.

mhd/godunov_fine.f90
tinecolman: Indeed. I've pushed a fix. Can you check again?
The indices +/- should exist no? Isn't that how the stencil is build?

PR #279, opened 2026-06-30
Add energy fix

tinecolman:
Add the mechanism to store the internal energy as a passive scalar at index ieint=nvar in uold.
This is used to avoid errors when computing the pressure from the total energy when the internal energy is much smaller than some of the other energies (typical high Mach flows).


rteyssier (commented):
I am confused: this is exactly what pressure_fix is already doing. What's the point of duplicating using another mechanism?

bcommerc:
Yes, but pressure_fix is only active in the hydro Godunov solver. Problems can still occur each time we need to compute the temperature from the total energy for some source terms (radiative transfer, chemistry). It is a bit different  compared to enew, which is recomputed form the total energy in set_unew. 

Maybe pressure_fix can be adapted then?

rteyssier:
Yes probably better to adapt pressure fix. Or use this new technique and drop pressure fix. Either way, we don't want to duplicate things in the code.

tinecolman:
Energy_fix doesn't add to the PdV source term in the current implementation, so we can't simply drop pressure_fix. We can combine energy_fix into pressure_fix, but then it is not backwards compatible in its current implementation due to the additional passive scalar added at compile-time.

PR #280, opened 2026-07-02
Merge the DICE reader patch in the main code

mtrebitsch:
Currently, reading DICE ICs relies on a patch that is hard to maintain, or to combine with other patches and modifications. This PR merges the content of the DICE patch in the main code, with hopefully minimal modifications.

Most of the modifications rely on using `dice_init` or specifying the IC filetype as `'dice'` in the namelist, with no additional compilation option. This passes the current tests (last I checked: https://github.com/mtrebitsch/ramses/actions/runs/28531261211).

The config file of the only test relying on DICE has been updated accordingly.

The code has not been tested for DICE setups in 1D and 2D, but I am not aware of such setup.

Minor comment (paging for example @tinecolman on this): I don't know if using `if` statements e.g. in `rho_fine.f90`, `particle_tree.f90`, and `add_list.f90` would lead to performance issue. If that is the case, maybe switching to `#ifdef` would be a reasonable compromise?

## PR Checklist

- [x] The pull request has a title and a description
- [x] The code covered by the pull request is documented


nbrucy:
Coool!

We need to remember updating the tutorials once this is merged.

hydro/init_flow_fine.f90
```diff
@@ -435,6 +464,37 @@ subroutine init_flow_fine(ilevel)
      end do
      ! End loop over grids
 
+  !-------------------------------------------------------
+  ! DICE initial conditions
+  !-------------------------------------------------------
+  elseif(filetype.eq.'dice') then ! TODO: dice stuff
```
tinecolman: Why is there a TODO here?

hydro/init_flow_fine.f90
```diff
@@ -435,6 +464,37 @@ subroutine init_flow_fine(ilevel)
...
+
+     do i=1,MAXGAL
```
tinecolman, line 473: This loop is MHD solver only?

hydro/init_flow_fine.f90
```diff
@@ -633,3 +693,870 @@ subroutine region_condinit(x,q,dx,nn)
...
+
+subroutine reset_uold(ilevel)
+  use amr_commons
+  use hydro_commons
+  use dice_commons
+  implicit none
+  integer::ilevel
+  !--------------------------------------------------------------------------
+  ! This routine sets array uold to zero before calling
+  ! the hydro scheme. uold is set to zero in virtual boundaries as well.
```
tinecolman: Maybe add that this is only for dice?

hydro/init_flow_fine.f90
tinecolman, line 1: There are a lot of additions here. I think some of the routines may fit better in other files?
Stuff like CIC should for sure for be in init_flow_fine but in rho_fine. Not sure what the best strategy is here: to try to distribute the routines to the corresponding files or to have 1 file with dice routines.

tinecolman:
Should the old patch directory be removed?

amr/init_refine.f90
```diff
@@ -136,3 +182,140 @@ end subroutine init_refine_2
 !################################################################
 !################################################################
 !################################################################
+subroutine kill_gas_part(ilevel)
```
tinecolman: this routine should go in the file where kill_tree is in pm/

pm/init_part.f90
```diff
@@ -86,6 +88,11 @@ subroutine init_part
 #ifdef OUTPUT_PARTICLE_POTENTIAL
   allocate(ptcl_phi(npartmax))
 #endif
+  ! Define up array (and maskp) for DICE ICs
+  allocate(up(npartmax))
+  if(ic_mask_ptype.gt.-1)then
+     allocate(maskp(npartmax))
+  endif
```
tinecolman, line 97: should only be allocated if dice

pm/rho_fine.f90
```diff
@@ -318,6 +330,198 @@ subroutine rho_from_current_level(ilevel)
...
+  use pm_commons
+  use hydro_commons
+  use poisson_commons
+  implicit none
+  integer::ilevel
+  !------------------------------------------------------------------
+  ! This routine computes the density field at level ilevel using
+  ! the CIC scheme from particles that are not entirely in
+  ! level ilevel (boundary particles).
+  ! Arrays flag1 and flag2 are used as temporary work space.
```
tinecolman: update description to make it clear this is for dice

tinecolman:
For performance, as long as the if is outside of a loop over cells/particles, it's fine.

tinecolman:
There are quite a few additional routines. Are they called only on init?

tinecolman (changes requested):

hydro/init_flow_fine.f90
mtrebitsch: Having 1 file with dice routines will probably be the "cleanest" (as in: less invasive), but then it's unclear in which folder to put it: it touches `amr/`, `hydro/`, and `pm/`. That being said, there is already a `amr/dice_commons.f90`, so maybe `amr/dice_routines.f90` would make sense?

mtrebitsch:
I'm working on fixes, and also adding the relevant documentation. That part is almost ready, except that I'm not sure I understand the `ic_mask_ivar` stuff (see here: https://github.com/mtrebitsch/ramses/blob/4135f9b6b6f93ca9512328abc5f510646c5e9194/hydro/init_flow_fine.f90#L784) :
```fortran
  do ind=1,twotondim
     iskip=ncoarse+(ind-1)*ngridmax
     do i=1,active(ilevel)%ngrid
        ! Initialisation of the refinement mask
        if((ic_mask_ivar.gt.0).and.(ivar_refine.gt.0).and.(ic_mask_ivar.le.nvar).and.(ivar_refine.le.nvar))then
     ! Switch to K/mu for ic_mask_ivar=ndim+2 case
           if(ic_mask_ivar.eq.ndim+2) then
        u = uold(active(ilevel)%igrid(i)+iskip,ic_mask_ivar)*scale_T2*(gamma-1)
           else
        u = uold(active(ilevel)%igrid(i)+iskip,ic_mask_ivar)
           endif
           if(ic_mask_ivar.gt.1)then
              u = u/uold(active(ilevel)%igrid(i)+iskip,1)
           endif
           if((u.ge.ic_mask_min).and.(u.le.ic_mask_max))then
              uold(active(ilevel)%igrid(i)+iskip,ivar_refine) = 1.0*uold(active(ilevel)%igrid(i)+iskip,1)
           endif
        endif
        e = 0d0
        do idim=1,ndim
           e = e+0.5*uold(active(ilevel)%igrid(i)+iskip,idim+1)**2/uold(active(ilevel)%igrid(i)+iskip,1)
        enddo
        uold(active(ilevel)%igrid(i)+iskip,ndim+2) = uold(active(ilevel)%igrid(i)+iskip,ndim+2)+e
     end do
  end do
```
My understanding is that if the value of the `ic_mask_ivar` field (divided by the density) is between two boundary values, the mask variable is set to the density. Is there any good reason to do that?

ROMAIN: My understanding is that for cosmo runs, GADGET has SPH (gas) particles only in the zoom region. In RAMSES, the zoom region is identified by a refinement mask. So if you have SPH particles, it means the deposited temperature will be non-zero, and this can be used as a way to identify the mask region.

tinecolman (commented):
Currently, some parts are in #if NDIM=3 blocks, while others are not. I propose we put  a single check on NDIM when the dice namelist is read and if it is not 3, then we exit with an error that dice is supported only in 3D.

Edit: Nevermind, there are NDIM blocks in init_dice that suggest that in principle it should work in 1D or 2D, so let's just leave it as is.

tinecolman (approved):
I think it's OK now.

nbrucy (changes requested):
Maybe we should remove the old patch file, no?

rteyssier (approved):

PR #282, opened 2026-07-03
Fix sink threshold accretion

tinecolman:
The accretion rate for threshold accretion shouldn't be multiplied by `(d/density)`. This creates too high accretion rates.

This bug was introduced when threshold accretion was put back into the code, but placed wrongly with respect to the new option to allow for volume-weighted bondi accretion.

Added a note to agn_acc_method that this parameter does not only apply to AGNs, but to sinks in general.


rteyssier:
Hi ramses-organisation,

I would not qualify this as a bug. This is just a mass-weighted accretion rate. The total accretion rate on the sink particle is the same, with the weight being either 1 (volume weighted) or d/density (mass weighted). Within the sink sphere, we have sum(d) = density here so it is properly normalized.

I suspect that this does not make sense for threshold accretion since in this case you accretion only d-d_threshold.

So this is not only a AGN only thing. For Bondi accretion in general for example this is a valid approach.

Romain
.

tinecolman:
@rteyssier Indeed, I've looked into it further since our last discussion. As you say, for Bondi accretion everything is fine. The problem is threshold accretion specific. The bug is that instead of accreting only the density above d_threshold, you weight even more by density, so accreting beyond d_threshold for the densest cells in the sink radius while accreting not enough from the less dense cells.

nbrucy (approved):

PR #283, opened 2026-07-03
fixes for induction scheme

tinecolman:
Fix missing (gamma-1) in computation of thermal energy.
Add call to  velocity_fine to correct solution at the end of set_uold.

Update the pressure solution for mhd/abc-flow test, which is now fixed to 1 (0 in reference solution).
Update the pressure solution for ponomarenko test, which is now smaller though not 0.

Found by @ulebreui


mhd/godunov_fine.f90
```diff
@@ -291,6 +291,9 @@ subroutine set_uold(ilevel)
      end if
   end do
 
+  ! Overwrite state if using induction scheme
+  if(ischeme == 1) call velocity_fine(ilevel)
```
rteyssier, line 295: It is already been called in mhd/courant_fine.f90. Seems redundant to call it here again.

tinecolman: It still does something though, because the pressure solution changes if I remove it.

tinecolman: The outputting is done before the call to velocity_fine in courant_fine by newdt_fine in amr_step.

rteyssier: Yes that is because of the position of the output dump.
I guess we can call it twice to really make sure nothing is left uncorrected.

rteyssier (approved):

PR #286, opened 2026-07-10
make initial pressure consistent with EOS for collapse condinit

tinecolman:
> [!WARNING]
>  This changes the solution for sims that use collapse condinit
>


rteyssier (commented):
I would disagree with this fix. The polytropic EOS is like a cooling function in a sense. It modifies the gas thermal pressure via the cooling_fine routine. If a user want to use initial conditions that do not sit at the equilibrium temperature of the cooling curve or at the polytropic temperature, that should be a possibility. If however the user want to have initial conditions sitting at the equilibrium cooling temperature or the polytropic temperature that is also an option that will be performed via a different set of initial conditions.

tinecolman:
I agree, that the user can choose whatever they want. Though shouldn't the default behavior be to be consistent with the other parameters set in the namelist?

tinecolman:
Screenshot from 2026-07-10 15-42-06

I was surprised at the difference it makes for the errors in the test solution.

rteyssier:
That is precisely my point: you do not want the initial conditions to be necessarily consistent with the EOS or the cooling function. Your initial conditions could be out-of-thermal equilibrium if that's what you want to study.

PR #288, opened 2026-07-15
Update to the restart mechanism for tests + consistency checks for output_params

tinecolman:
## Improvements to the mechanism that is testing restarts

1) Rewrite the restart script to allow for a wider range of output parameters in test namelists. This should fix typical issues encountered when adding new tests.

2) Add a test config option that will run individual tests using the restart mechanism by default. This option is added as a second line in config.txt, where the compilation parameters are specified. This option has been enabled for most tests.

Restarts are now tested **by default**.

## Improvements for output_params handling

Having a wider variety of options to tests restarts revealed some issues with combinations of output namelist params. Particularly, the output_params of the  sfvirialtrue-sfmodel0 created a nonsense tout list. This resulted in no output being written at tend. To avoid this behavior, I've added more consistency checks on the use of `tout`, `aout`, `tend` and `aout`:
* tend/aend should be >= to the last tout/aout given
* the values of tout/aout should be increasing

Since noutput is calculated automatically (introduced in a previous PR), the namelist parameter `noutput` is now deprecated, to avoid users providing a wrong value. If provided, it will simply be reset and print a warning in the log.

 > [!WARNING]
 `noutput` is now deprecated. Setting it in the namelist will not do anything.

The dice patch has been updated, since the SF/FB tests were affected. The sfvirialtrue-sfmodel0 test has been updated to properly use delta_tout for testing purposes.

## Other

* Fixed a formatting bug in the test suite log where at the end everything would be printed on one super long line. This caused python errors thrown by visu_ramses or the test plotting scripts to become difficult to read.

* Fixed the tracer particle test which had very large tolerances and no check on the output time. This caused the crash at restart to go undetected and wrongly announced the test as passed.


tinecolman:
Currently, not all tests pass when using the restart mechanism. This is why we do not set restart as default for all tests.

tests/run_with_restart.py
```diff
@@ -5,168 +5,274 @@
...
 
-Usage:
-python run_with_restart.py -s <step> -t <test_name>
+The idea of a restart test is to reproduce, in two chunks joined by a restart,
+exactly the same final output(s) that a single uninterrupted run would produce.
+The test suite invokes this script in three steps around two runs of the code:
+
+    python run_with_restart.py -s 1 -t <test_name>   # shorten the run
+    <run the code>                                   # run 1 -> stops mid-way
+    python run_with_restart.py -s 2 -t <test_name>   # restore + set nrestart
```
nbrucy: What is the meaning of "restore" in this context

tests/run_with_restart.py
```diff
@@ -5,168 +5,274 @@
...
+    candidates = set()
+    if var == "time":
+        explicit, delta = tout, delta_tout
     else:
-        print("ERROR: noutput found but tout or aout not found in output_params")
+        explicit, delta = aout, delta_aout
+    for x in explicit:
+        if 0 < x < end:
+            candidates.add(x)
+    if delta is not None and 0 < delta < FOUTPUT_OFF:
```
nbrucy: I think `FOUTPUT_OFF` this should be replaced by `end`

tests/run_with_restart.py
```diff
@@ -5,168 +5,274 @@
...
+        k = 1
+        while k * delta < end and k < FOUTPUT_OFF:
```
nbrucy: Why this `k < FOUTPUT_OFF` check?

tests/run_with_restart.py
```diff
@@ -5,168 +5,274 @@
...
-        else:
-            nml["output_params"]["aout"] = aout * factor
+    tout = _as_list(op.get("tout"))
+    aout = _as_list(op.get("aout"))
+    tend = op.get("tend", 0) or 0
+    aend = op.get("aend", 0) or 0
+    delta_tout = op.get("delta_tout")
+    delta_aout = op.get("delta_aout")
+    foutput = op.get("foutput")
+    foutput_active = foutput is not None and 0 < foutput < FOUTPUT_OFF
```
nbrucy: Problematic. That does not guarantee there will be a foutput triggered output before the end of the simulation.

tests/run_with_restart.py
```diff
@@ -5,168 +5,274 @@
...
+    """Pick where run 1 should stop and how many extra outputs that creates."""
+    candidates = sched["candidates"]
+    if candidates:
+        # Stop on the middle scheduled output -> continuous numbering.
+        split = candidates[(len(candidates) - 1) // 2]
+        offset = 0
+    else:
+        # No intermediate reference output: stop at the mid-point.
+        split = sched["end"] / 2.0
+        # With foutput active every step is an output, so the mid-point dump is
```
nbrucy: Wrong actually. Depends of the value of foutput

tests/run_with_restart.py
```diff
@@ -5,168 +5,274 @@
...
+        op["aend"] = split
+    elif mech in ("tout_list", "aout_list"):
+        key = "tout" if mech == "tout_list" else "aout"
+        if single:
+            op[key] = split
+            if sched["has_noutput"]:
+                op["noutput"] = 1
+        else:
+            new = sorted(x for x in _as_list(op.get(key)) if x <= split)
+            if split not in new:
```
nbrucy: This should never happen

tests/run_with_restart.py
```diff
@@ -5,168 +5,274 @@
...
+    for k in range(extra + 1, last + 1):
+        src = f"output_{k:05d}"
+        dst = f"output_{k - 1:05d}"
+        os.rename(src, dst)
+        for fname in os.listdir(dst):
+            # Replace the (first) output-number token; leave the .outNNNNN cpu
+            # index untouched.
+            new = re.sub(f"{k:05d}", f"{k - 1:05d}", fname, count=1)
+            if new != fname:
+                os.rename(os.path.join(dst, fname), os.path.join(dst, new))
```
nbrucy: Uselessly generic. As stated in the comment, in that case there will be 3 only outputs, so the we could use the old version of the code.

nbrucy (changes requested):
I'm not sure the foutput detectecion works as intended. I suggest to remove it and keep forbiding to set foutput in tests setup (as it is very hard to predict when the outputs will be created). An alternative is to just allow for foutput=1.

tinecolman:
> I'm not sure the foutput detectecion works as intended. I suggest to remove it and keep forbiding to set foutput in tests setup (as it is very hard to predict when the outputs will be created). An alternative is to just allow for foutput=1.

Yes, foutput is still not working. The only way to use it would be with a max number of steps in run params.

tests/run_with_restart.py
tinecolman: restore the original output time

tests/run_with_restart.py
tinecolman: yes

tests/run_with_restart.py
tinecolman: True

tinecolman:
I've refactored the sketchy parts of the script. And found a bunch of other issues...

tests/run_with_restart.py
```diff
@@ -10,133 +10,232 @@
...
+  -s <step> is the step to run (1, 2 or 3)
+  -t <test_name> is the test name
+
+Supported &OUTPUT_PARAMS options
+--------------------------------
+The output schedule can be driven by any combination of:
+  tout / aout            explicit output times / expansion factors (scalar or list)
+  tend / aend            end time / end expansion factor (appended as final tout/aout)
+  delta_tout/delta_aout  periodic outputs in time / expansion factor
+  noutput                number of predefined outputs (recomputed by RAMSES)
```
nbrucy: Not relevant anymore

tests/run_with_restart.py
```diff
@@ -10,133 +10,232 @@
...
-        modifying it.
+    Reconstruct, from the &OUTPUT_PARAMS block, the information needed to split
+    the run. Mirrors the termination logic of read_output_params/update_time.
+
+    Returns a dict with:
+      var    : 'time' or 'expansion' -- the controlling variable
+      end    : value of that variable at which the run terminates
+      end_type   : how the end is imposed ('tend'/'aend'/'tout'/'aout')
+      candidates : sorted intermediate reference outputs strictly in (0, end)
+      has_noutput    : whether noutput is written explicitly in the namelist
```
nbrucy: not relevant anymore

tests/run_with_restart.py
```diff
@@ -10,133 +10,232 @@
...
+        k = 1
+        while k * delta < end:
+            candidates.add(k * delta)
+            k += 1
+
+    return dict(
+        end=end,
+        end_type=end_type,
+        candidates=sorted(candidates),
+        has_noutput=("noutput" in output_params),
```
nbrucy: not needed

tests/run_with_restart.py
```diff
@@ -10,133 +10,232 @@
...
+        updated_params = _as_list(output_params[end_type])
+        if split not in updated_params:
+            updated_params.append(split)
+        updated_params = sorted(updated_params)
+        # cut the output list at the new end time
+        i_end = updated_params.index(split)
+        updated_params = updated_params[0:i_end+1]
+        output_params[end_type] = updated_params
+        # update noutput if set
+        if schedule["has_noutput"]:
```
nbrucy: ignored

nbrucy (changes requested):
I think it's good now, except that there are still reference to noutput. I suggest also to remove all reference to noutput in all the test's namelist, since these namelists are sometimes used as the starting point to build more complex simulation.

tinecolman:
This system revealed that restarts with TURB were broken by PR #277 . This is fixed by PR #289

nbrucy (approved):

nbrucy:
The CI fails because ramses.cnrs.fr is out, so it should come back in a while

PR #289, opened 2026-07-21
Bug fixes in the TURB module

tinecolman:
## Summary

This PR fixes restarts for tests that uses the TURB module. These have been broken by PR #277. The cause was a duplication of the interpolation formula, which has now been removed by added a new routine `turb_interpolate_now`. Tracking down this bug revealed a few other issues which are also fixed here. All bugs are detailed below.

Tests have been added to verify the behavior of type 2 and type 3 turbulence. Turbulence tests now include a figure that shows the evolution of the rms and kinetic energy. Reference solutions of all tests using TURB had to be regenerated because the turbulence field changed.

A section describing what each of the three types actually does has been added to the wiki.

Co-authored by Claude Code, which helped to find the bugs and write documentation.

### 1. Type 1: restarts diverge from the continuous run

`#277` corrected the time-interpolation weights in `turb_check_time` but the same expression also existed in `init_turb`, where it was left inverted. The value computed by `init_turb` is the one actually used on the first step, since `calc_turb_forcing` (`amr_step`) runs before `update_time`/`turb_check_time`.

On a fresh start the time fraction is 0, so the two formulas agree by accident. After a restart `t` falls part way through the interval, the weights are genuinely inverted, and the first post-restart step is forced with a mis-weighted field. Turbulence then amplifies the difference well past the test tolerance.

Before #277 both copies were wrong in the same way, so restarts were self-consistent; #277 made them inconsistent. Fixed by correcting `init_turb`.

### 2. Type 3 with MPI: ranks disagree on the initial field

The `turb_type == 3` clock adjustment in `init_turb` mutated `turb_last_time` *after* `mpi_share_turb_fields` had already broadcast it, and non-root ranks returned from the routine before ever reaching it. Rank 1 therefore selected
`afield_next` while every other rank selected `afield_last`.

For decaying turbulence that field *is* the initial velocity field, so a parallel run started from initial conditions that disagreed across domains. Type 1 was unaffected (no such adjustment). Fixed by restructuring `init_turb` so all ranks follow one code path and the broadcast happens at a single rendezvous.

### 3. Type 2: wrong rms after a restart

`init_turb` applied time interpolation to a field that is static by definition. On a fresh start the weights collapse to a single field, but after a restart `t` lies mid-interval, so the "fixed" field became a convex blend of two independent realisations — and such a blend has an rms *below* `turb_rms`.

This is the defect that `read_turb_params` documented and rejected outright with *"Turbulence type 2 is bugged. Please select 1 instead."* Fixed by selecting a single field and never interpolating it for type 2. The namelist rejection has been removed, so type 2 is usable again.

### 4. Types 2 and 3: forcing amplitude systematically off target

A single draw of the Ornstein-Uhlenbeck process has an rms a few percent away from `sqrt(ndim)*turb_rms`. For type 1 that scatter is intended and averages out over a run. For type 2 the field is frozen, and for type 3 it is used once as the initial velocity, so in both cases the deviation of that one draw becomes a systematic error for the entire simulation. Both are now renormalised at startup so the amplitude is exact. Type 1 is deliberately left untouched.

### 5. `instant_turb`: field schedule advanced one interval too far

After the spin-up loop, two lines advanced `turb_last_time`/`turb_next_time` by one `turb_dt` without generating a corresponding field. The loop already leaves the clock at the correct bracket, so this pushed `t` outside the interval.

Two consequences: the time fraction came out at exactly `-1`, so the initial `afield_now` was the extrapolation `2*afield_last - afield_next` rather than a field on the segment; and every subsequent field update fired one interval late.

This affects every type-1 run using `instant_turb`, which is on by default, so it applies to all existing turbulence tests. Confirmed by instrumenting `turb_tfrac` (`-1.0` with `instant_turb`, `0.0` without) and by the reported rms series shifting forward by exactly one entry once fixed.

### 6. Single-precision builds: PRNG returned garbage

The non-`DOUBLE_PRECISION` branch of `kiss64_double` reinterpreted the crafted int64 bit patterns using a *real4* mold instead of real8, and omitted the `- 1.0` that maps `[1,2)` onto `[0,1)`. It returned values such as `1.4e-40`, `1.875`, `2.8e-40` instead of uniform variates.

Only compiled for `NPRE=4` (or `NPRE` unset), so double-precision builds — and hence every reference solution — are unaffected. Latent, but silent.


rteyssier (approved):

doc/wiki/TurbulenceDriving.md
```diff
@@ -71,13 +71,102 @@ fraction $\chi$.
 |:----------------------|:------------ |:------------- |:---------| :------------------------ |
 | `turb`                | `boolean`    | `.false.`     |          | Turn on or off driving
 | `turb_seed`           | `integer`    | `-1`          |          | Random number generator seed. -1 = random
-| `turb_type`           | `integer`    | `1`           |          | How the driving changes over time. 1=driven evolving, 3=decaying
-| `instant_turb`        | `boolean`    | `.true.`      |          | Generate initial turbulence before start
+| `turb_type`           | `integer`    | `1`           |          | How the driving changes over time. 1=driven evolving, 2=driven fixed, 3=decaying. See the "Turbulence types" section below
+| `instant_turb`        | `boolean`    | `.true.`      |          | Evolve the field to saturation before the run starts. Recommended for `turb_type=1`. See the "Turbulence types" section below
```
nbrucy, line 75: If it's recommended for type 1, maybe put it as true in the driving test

doc/wiki/TurbulenceDriving.md
```diff
@@ -71,13 +71,102 @@ fraction $\chi$.
...
+above, being advanced every $dt = T/N_{dt}$ and linearly interpolated in between,
+and is applied as an acceleration at every timestep. The timestep is capped at
+$T/N_{dt}$ so that no step skips over a field update.
+
+The rms of the field fluctuates by a few percent about
+$\sqrt{N_{dim}}\,f_{\mathrm{rms}}$ from one update to the next. This is intended:
+$f_{\mathrm{rms}}$ sets the *time-averaged* amplitude, and the scatter averages
+out over a run.
+
+An evolving field grown from rest starts at only $\sqrt{1-e^{-2dt/T}}\approx14\%$
```
nbrucy:
```suggestion
An evolving acceleration field grown from rest starts at only $\sqrt{1-e^{-2dt/T}}\approx14\%$
```

doc/wiki/TurbulenceDriving.md
```diff
@@ -71,13 +71,102 @@ fraction $\chi$.
...
+of its saturated amplitude and needs about one autocorrelation time $T$ to build
+up. The `instant_turb` option (on by default) evolves the field for several $T$
+before the run starts, so that it begins already saturated. Leaving it on is
+strongly recommended for `turb_type=1`, especially for runs shorter than or
+comparable to $T$, which would otherwise spend most of their length in the
+spin-up transient. It has no useful effect for the other two types.
+
+### `turb_type=2` — fixed driving ###
+
+One field is generated at startup and then held, unchanged, for the whole run.
```
nbrucy:
```suggestion
One acceleration field is generated at startup and then held, unchanged, for the whole run.
```

tests/turb/driving-decaying/driving-decaying.nml
```diff
@@ -0,0 +1,76 @@
+This is the RAMSES parameter file for the decaying turbulence test.
```
nbrucy, line 1: The name of this test is a bit confusing.

nbrucy (changes requested):

doc/wiki/TurbulenceDriving.md
tinecolman: It's on by default

tests/turb/driving-decaying/driving-decaying.nml
tinecolman: That's because the option is confusion. I change it in the next PR.

nbrucy (approved):

PR #291, opened 2026-07-22
[TURB] make the initial velocity field a standalone feature + replace `turb_type`

tinecolman:
## Summary

The parameters describing the turbulence have been overhauled, allowing for new combinations.

`turb`-> `driven_turb` (simple renaming)
`turb_type=3` -> `initial_turb`, `initial_turb_vrms`, `initial_turb_spectrum`, `initial_turb_comp_frac`
`turb_type=1`-> `turb_evolving=.true.`
`turb_type=2`-> `turb_evolving=.false.`

NEW `turb_exact_rms`: always rescale rms to namelist value.

This change is made backward compatible with check in read_turb_params.

## Motivation

`turb_type` bundled three unrelated things behind one integer:

- **1** and **2** were two genuine driving modes (evolving vs. static field).
- **3** was not a driving mode at all. It applied the turbulent field *once*, as an initial velocity, then switched off every forcing path for the rest of the  run. It only lived in the driving module because it reused the field generator.

That made `turb_type=3` hard to reason about and hard to combine with anything. It also meant `turb_rms` silently changed meaning between types (an acceleration for 1 and 2, a velocity for 3), and it scattered `turb_type /= 3` guards through `amr_step`, `courant_fine` and `newdt_fine`.

This PR splits the "initial" turbulence from the "driving" turbulence.

### The initial velocity field is now its own feature

`turb_type=3` has been replaced by `initial_turb`. This starts a run from a turbulent velocity field, **independent of whether driving is on**. That enables three combinations, one of which was previously impossible:

| `driven_turb` | `initial_turb` | Result |
|:---|:---|:---|
| `.true.`  | `.false.` | Driving only: gas starts at rest and is stirred |
| `.false.` | `.true.`  | Decaying: gas starts turbulent, never driven |
| `.true.`  | `.true.`  | Driven, starting from an already turbulent field ← new |

Implementation notes:

- **Its own entry point.** `add_turb_init_velocity` adds the velocity and updates the total energy directly. Previously this went through `synchro_hydro_fine(ilevel, 1.0_dp, 2)` — passing `1.0` as a *timestep* so that an acceleration would be integrated into a velocity over exactly one code time unit. `synchro_hydro_fine` is now purely about forces again.
- **Its own field.** The initial field is an independent draw, not a rescaled copy of the driving field, so the two are uncorrelated. It also has its own spectrum (`initial_turb_spectrum`) and compressive fraction (`initial_turb_comp_frac`), so you can e.g. start from broadband turbulence and drive a single mode.
- **Its amplitude is a velocity.** `initial_turb_vrms` is the resulting |v|_rms directly, not something you divide by sqrt(ndim) first.

### `turb_type` 1/2 becomes a boolean

The remaining `turb_type`1 and 2 have been replaced by `turb_evolving` (default `.true.`):

- `.true.` — Ornstein-Uhlenbeck evolution, interpolated between updates (previously turb_type=1)
- `.false.` — one static field held for the whole run (previously turb_type=2)

### New option: `turb_exact_rms`

Sets the injected rms to exactly sqrt(ndim)·`turb_rms` instead of letting it follow the random draw. This makes `turb_evolving` and `turb_exact_rms` two orthogonal axes describing the driving pattern and amplitude separately. This exposes a mode that wasn't reachable before: a pattern that keeps decorrelating at constant injected power.

## Tests

Renamed so the names match the parameters they exercise:

- `turb/driving` → `turb/driving-evolving`
- `turb/driving-fixed` → `turb/driving-static`
- `turb/driving-decaying` → `turb/decaying` (no driving involved any more)

New:

- `turb/driving-parabolic` — the `parabolic` spectrum, which is the module default and had no coverage
- `turb/driving-initial` — the only test with `initial_turb` and `driven_turb` both on, covering their interaction. Broadband turbulent initial field, then a static fully compressive push on a single mode; the driving rms must stay pinned on target, i.e. undisturbed by the initial field.


rteyssier (approved):

bin/Makefile
```diff
@@ -179,8 +179,9 @@ ifeq ($(GRACKLE),1)
    LIBS_OBJ     = -I$(HOME)/local/include -DCONFIG_BFLOAT_8 -DH5_USE_16_API -fPIC
 endif
 ifeq ($(USE_TURB),1)
-   LIBS_TURB = -L$(HOME)/local/lib -lfftw3
-   LIBS_OBJ_TURB = -I$(HOME)/local/include -lfftw3
+   FFTW_DIR ?= $(HOME)/local
```
nbrucy, line 190: Why a safe asignement here? What the intention to use that as a defaultt value?
In that case the syntax is  `${FFTW_DIR:=$(HOME)/local}`

doc/wiki/TurbulenceDriving.md
```diff
@@ -71,102 +71,134 @@ fraction $\chi$.
 |:----------------------|:------------ |:------------- |:---------| :------------------------ |
 | `turb`                | `boolean`    | `.false.`     |          | Turn on or off driving
```
nbrucy: Following the logic, shouldn't this be renamed `turb_driving`?

hydro/synchro_hydro_fine.f90
```diff
@@ -13,7 +13,9 @@ subroutine synchro_hydro_fine(ilevel,dteff,which_force)
   real(dp)::dteff
   integer::which_force !gravity=1, turbulence=2
   !-------------------------------------------------------------------
-  ! Update velocity  from gravitational acceleration
+  ! Update velocity as v = v + force*dteff, where the force is either the
```
nbrucy:
```suggestion
  ! Update velocity as v = v + force*dteff, where the "force" is either the
```

(because it's indeed not a force but an acceleration)

tests/turb/decaying/decaying.nml
```diff
@@ -0,0 +1,74 @@
+This is the RAMSES parameter file for the decaying turbulence test.
```
nbrucy:
```suggestion
This is the RAMSES parameter file for the decaying turbulence test, using the built-in turbulence module.
```

nbrucy (changes requested):
Good for me, just a few thing to check before merging.

doc/wiki/TurbulenceDriving.md
tinecolman: Yes. But I didn't want to change the existing name for backwards compatibility reasons. But it can be changed if you think it is better?

hydro/synchro_hydro_fine.f90
tinecolman: I can just change "force" to "acceleration"

doc/wiki/TurbulenceDriving.md
nbrucy: Maybe it's better to change, but continue to accept turb in the namelist for backwards_compatibility (and print a depreciation warning).

tinecolman: Iv' replaced it by `driven_turb`, symmetrical to ìnitial_turb`

bin/Makefile
tinecolman: No, the idea is to set it to the default home path only if FFTW has not yet been set to something else. For example, when doing a module load fftw, it _should_ be already set by the cluster if properly configured.

nbrucy (approved):

PR #296, opened 2026-07-28
Add option to compile with the new LLVM-based Intel Compiler ifx

tinecolman:
### Summary
Set `COMPILER=INTELX` to compile with the new intel compiler. The old ifort compiler remains available under the regular `COMPILER=INTEL`.

### Background
The ifort compiler has been discontinued in 2024 and has been replaced by ifx.
https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2024-2/overview.html

This means that the old ifort may no longer be available on clusters in the future. This seems to be already the case for MeluXina.

While the old versions of ifx did not support all compiler flags in the RAMSES Makefile, the newer versions do. Users are advised to module load the most recent version, for example on MareNostrum:
module load intel/2025.2
module load impi/2021.16

Switching to ifx may impact performance (in good or bad ways). On marenostrum, the performance is similar for sedov and cosmo benchmarks when using MPI. However, others have reported performance degradation. It is adviced to benchmark before using it for production runs.


nbrucy:
Is ifx available in all clusters? I'm afraid the change may generate compilation issues for people running on older clusters ...

tinecolman:
I have no way of knowing the intel version on very cluster. But I think the default in the Makefile shouldn't be a deprecated version. Currently it may generate compilation issues on new clusters.
Also I found a bug in ifort with openmp...

rteyssier (approved):
May be we could use ifx as default for default COMPILER=INTEL and use the old version for COMPILER=INTEL_OLD in a second compiler block in the Makefile?

tinecolman:
After running more tests, I found that ifx is significantly slower when using openmp on the sedov benchmark. It seems others also report performance issues. So I agree we need to keep both.

PR #298, opened 2026-08-10
remove dummy dt_new argument

nbrucy:
This micro PR fixes this warning: 

```
../rt/rt_spectra.f90:677:61: warning: type of ‘heat_unresolved_hii_regions’ does not match original declaration [-Wlto-type-mismatch]
  677 |        call heat_unresolved_HII_regions(ilevel,dtnew(ilevel))
      |                                                             ^
../rt/rt_cooling_module.f90:1260:38: note: type mismatch in parameter 2
 1260 | SUBROUTINE heat_unresolved_HII_regions(ilevel)
      |                                      ^
../rt/rt_cooling_module.f90:1260:38: note: ‘heat_unresolved_hii_regions’ was previously declared here
```

### DIFF
--- a/rt/rt_spectra.f90
+++ b/rt/rt_spectra.f90
@@ -674,7 +674,7 @@ SUBROUTINE star_RT_feedback(ilevel, dt)
   ! End loop over cpus
 
   if(heat_unresolved_HII.gt.0) &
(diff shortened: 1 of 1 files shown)


tinecolman (approved):
dtnew is an amr_commons, no need to pass it in any case

PR #299, opened 2026-08-10
update automatic tests to use 4 cpus

tinecolman:
The number of cores on ubuntu-latest has been updated from 2 to 4.


tinecolman:
I'm checking first whether the CI actually runs with 4 cpus and which tests are failing.

tinecolman:
The MC tracer test fails because of the random number generator which is initialized per MPI ranks.
The classical tracer test also fails, even though it does not use random numbers. To be investigated.
The ponomarenko test fails because of some deeper issue, see PR #300 .

tinecolman:
This isn't actually faster. Probably because there are only 2 _physical_ cores.

Issue #305, opened 2026-08-15
Dice IC not adapted to LIGHT_MPI


tinecolman:
up/maskp never converted in virtual_tree_fine

PR #310, opened 2026-08-19
Fix broken compilation of LIGHT MPI since dice patch merge

tinecolman:
Add the `LIGHT_MPI_COMM` preprocessor switch pattern for `reception` and `emission` arrays to dice related blocks. This restores compilation for LIGHT MPI.


tinecolman:
This solves issue #305

nbrucy (approved):

PR #312, opened 2026-08-19
[WIP] Series of sink-related improvements

tinecolman:
## Summary

This PR fixes/improves a variety of things related to sinks and how they are tested

## Improvements for sinks

1) **Radiation feedback deposited using CIC instead of NGP**
   The NGP scheme will simply deposit all radiation in the nearest cell for each cloud particle. With CIC, the radiation will be distributed among the 2**ndim cells overlapping with the particle, weigthed be the overlap value. This is a smoother approximation.
   In particular, this solve an issue which occurs when a sink is sitting at a cell boundary (as is the case in some of the sink tests). Since clouds particles are spawned on a lattice with spacing dxmin/2, most cloud particles will also sit at cell boundaries. In that case, NGP does a truncation to decide in which cell to dump the radiation. This truncation is alway in the same direction. 
   Two cloud particles mirrored about the sink then deposit on the same side, offsetting the whole radiation source by half a cell. Which in turn creates a spurious net force on the sink. This could be measured in the sink/stellar-HII test, where the symmetry of the setup dictates that vz_sink=0. Using CIC (which is the standard in the code anyway), removes this directional bias and restores vz to something close to 0.
   Note that by default not all CIC cells exist, since some can be on a different refinement level. In that case the radiation is lost. However, this problem is solved by setting sink_refine=.true., which guaranties that the area of the sink is refined and all CIC cells exist. A warning for this has been added.

2) **Increased precision of the sink and stellar output csv file**
   The output file now outputs the full double instead of truncating it after 10 decimals. Before, information was lost upon restart. 
   Additionally, force fortran to always write the E for scientific notation.
   This change **is backwards compatible** due to the comma separator in the csv file. Existing outputs remain restartable.

## General test suite improvements

**Issue:** The relative errors for variables that should have been zero due to symmetry reasons, but where not because of the limited accuracy and non-symmetry-conserving properties of the poisson solver would be very large. Basically, the test suite would compare a reference solution which contained only numerical noise with another solution which was also numerical noise of similar magnitude.
**Symptom:** This seems to be at the base of certain tests not passing when running with a different nproc or with the restart mechanism. The affected tests are all setups with a certain symmetry which have gravity: hydro/isothermal and barotrop which are 1D collapsing to the center, and sink/center-SN and stellar-HII which have sinks in the x-y plane of a uniform box.
**Solution:** Manually reset the references for variables that should be zero to zero. To accommodate this, a change of the behavior of the visu_ramses error computation for solutions that should be 0 has been implemented: when the reference is 0 we compute the absolute error instead of the relative error.

## Improvements to sink tests

* Added missing sink variables in the reference solutions of center-SN, stellar-HII and stellar-spawn tests.
* Added a description to the setup for stellar-HII, with particular attention to the consequence of the symmertry.
* TODO reference to it for center-SN
* Due to the symmetry in the setup of stellar-HII and center-SN, sink_vz should be 0. This physically motivated solution has been


pm/output_sink.f90
```diff
@@ -9,14 +9,21 @@ subroutine output_sink_csv(filename)
...
 
   fileloc=TRIM(filename)
-  open(unit=123,file=TRIM(fileloc),form='formatted',status='replace', recl=500)
+  ! recl must hold the whole record: I10 + 21*(A1+ES25.16) + A1 + I10 = 567.
+  ! ES25.16 is 17 significant digits, which is what an IEEE double needs to
+  ! restart exactly.
+  ! The explicit E3 matters. Without it Fortran drops the 'E' when the exponent
+  ! needs three digits, writing e.g. 1.2345678901234567-123. Fortran reads that back
+  ! correctly, but numpy cannot, so visu_ramses turns any such value into NaN
+  ! when the test suite reads the file. E3 always emits the E.
```
rteyssier, line 18: These are typical Claude comment style. I would remove.

pm/sink_particle.f90
```diff
@@ -2695,6 +2698,24 @@ subroutine read_sink_params()
...
+#ifdef RT
+  ! Radiation from the sink clouds is deposited with a CIC stencil at the level
+  ! each cloud particle is attached to, and sink_RT_feedback runs after
+  ! kill_tree_fine, so a stencil volume reaching into a coarser neighbour is
+  ! dropped with no other level's call to pick it up. sink_refine is what makes
+  ! that impossible: it refines all eight CIC cells of every cloud particle up to
+  ! nlevelmax_sink. Without it, a sink near a refinement boundary silently emits
+  ! less than its stellar objects should. See the header of
+  ! sink_RT_vsweep_stellar in pm/sink_rt_feedback.f90.
+  if (rt_sink .and. .not.sink_refine)then
```
rteyssier, line 2710: Same comment on the Claude style comment. Way too verbose.

pm/sink_rt_feedback.f90
```diff
@@ -175,11 +175,54 @@ END SUBROUTINE gather_ioni_flux
...
+! With sink_refine=.false. that guarantee is gone and a sink near a refinement
+! boundary will emit slightly less than its stellar objects should. Use
+! sink_refine.
+!
+! Note this covers the coarser direction only. A volume falling in a neighbour
+! that is refined FINER than ilevel is skipped as well, and that is correct:
+! rt_upload_fine re-imposes a split cell's value from its children, so photons
+! put there would be erased. Those photons are lost, which is a real if much
+! smaller effect, and needs gas criteria to refine past nlevelmax_sink next to a
+! cloud for it to arise at all.
```
rteyssier, line 225: This is way too long...

rteyssier (changes requested):
This is a good commit. All sink related feedback should be CIC based.
Try and avoid verbose comments a la Claude.

tinecolman:
> This is a good commit. All sink related feedback should be CIC based. Try and avoid verbose comments a la Claude.

I know, that's one of the reasons why it is still in draft.
The other being that the stuff in PR #313 also affects the sink test solutions.
And the RT compilation being broken...

PR #313, opened 2026-08-20
More control over poisson multigrid solver safe_mode

tinecolman:
## Summary
1) Output the state of `safe_mode` to disk, so it can be loaded on restart. This fixes the barotrope and isothermal tests failing on restart, as it was silently reset.
2) Adds a mechanism to control after how many solves 'safe_mode' should be reset, using a new namelist parameter `mg_safe_mode_reset`
   * `mg_safe_mode_reset=0`: never reset (old behavior, at least until restart)
   * `mg_safe_mode_reset=X` with X>=1: reset every X solves
3) Add a new namelist parameter 'mg_maxiter' to allow control over the max number of iterations for the multigrid solver, which was previously hard-coded.

## Implementation

An additional array `safe_mode_countdown(ilevel)` is allocated to keep track of how many solves we still want to do in safe_mode. The counter is set to  `mg_safe_mode_reset` when safe_mode is switched on and is decreased at the start of every solve. When the countdown reaches 0, the code resets `safe_mode=.false.`.

The state of safe_mode is outputted to a new file `safe_mode_NOUT.txt`. This file contains `safe_mode` and `safe_mode_countdown`, so the state can be restored upon restart. For example:
```
# Multigrid safe mode state
levelmax    8
mg_safe_mode_reset        0
# level  safe_mode  solves_left
    1        F           0
    2        F           0
    3        F           0
    4        F           0
    5        F           0
    6        F           0
    7        T           0
    8        F           0
```

## Why this fixes the barotrop and isothermal tests

This was caused by the poisson multigrid solver flag `safe_mode(ilevel)` being silently reset on restart.

### The problem
In these tests, the multigrid solver fails to converge for the two deepest AMR levels on initial conditions. Consequently, the code was setting  `safe_mode=.true.` on these levels for **the remainder of the simulation**.
However, this flag is not outputted and so upon restart it is reset to its default which is `.false.`. This was apparently the intended but undocumented behavior. There are no further convergence issues after the initialisation. So after restart, the affected AMR levels run with `safe_mode=.false.`. This results in a different solution for the gravitational potential and thus a different test solution.

### The fix
Dump `safe_mode(ilevel)` into the output and read it on restart. With the restored safe_mode state, the reference solution is recovered exactly for the hydro/barotrop and isothermal tests.

### Other changed for these tests

* At the start of this work, both reference solutions have been regenerated on 2 cpus
* read/write_conservative have been added to their namelists. This adds the `momentum_x` and `total_energy` fields to their reference solutions, without changing the other variables.

Credit to Claude Code for helping me find the cause of this issue.


poisson/multigrid_fine_commons.f90
```diff
@@ -191,6 +190,7 @@ subroutine multigrid_fine(ilevel,icount)
 
    iter = 0
    err = 1.0d0
+   safe_mode(ilevel) = .false.  ! Reset safe_mode
```
rteyssier: This will force the poisson solver to fail every time. I am not sure I see the point of that change.

rteyssier (changes requested):
This is not a proper fix for this problem. I would rather output safe_mode(level) to disk so that it is read back at restart.

poisson/multigrid_fine_commons.f90
tinecolman: What do you mean fail every time? It only fails to converge once, in the beginning when the initial guess is bad. After that it converges fine:

```
 Initial mesh structure
 Level  1 has          1 grids (       0,       1,       0,)
...
 Level 16 has        348 grids (     174,     174,     174,)
 Starting time integration
   ==> Level=    5 Step=    6 Error= 5.033E-05
   ==> Level=    6 Step=    5 Error= 4.597E-05
   ==> Level=    7 Step=    5 Error= 4.636E-05
   ==> Level=    8 Step=    5 Error= 4.638E-05
   ==> Level=    9 Step=    5 Error= 8.583E-05
   ==> Level=   10 Step=    5 Error= 4.132E-05
   ==> Level=   11 Step=    5 Error= 4.329E-05
   ==> Level=   12 Step=    5 Error= 4.703E-05
   ==> Level=   13 Step=    5 Error= 4.637E-05
   ==> Level=   14 Step=   12 Error= 8.749E-05
   ==> Level=   15 Step=   20 Error= 2.223E+00
 WARN: Fine multigrid Poisson failed to converge...
   ==> Level=   16 Step=   20 Error= 1.813E+00
 WARN: Fine multigrid Poisson failed to converge...
 Fine step=      0 t= 0.00000E+00 dt= 1.906E-05 a= 1.000E+00 mem=14.7% 
 Time elapsed since last coarse step:    0.02 s       26.11 mus/pt       26.11 mus/pt (av)
 Used memory:     30.4 MB
 Total running time:   4.99999970E-02 s
   ==> Level=    5 Step=    6 Error= 5.033E-05
   ==> Level=    6 Step=    5 Error= 4.597E-05
   ==> Level=    7 Step=    5 Error= 4.636E-05
   ==> Level=    8 Step=    5 Error= 4.638E-05
   ==> Level=    9 Step=    5 Error= 8.598E-05
   ==> Level=   10 Step=    5 Error= 4.132E-05
   ==> Level=   11 Step=    5 Error= 4.328E-05
   ==> Level=   12 Step=    5 Error= 4.704E-05
   ==> Level=   13 Step=    5 Error= 4.637E-05
   ==> Level=   14 Step=    5 Error= 4.621E-05
   ==> Level=   15 Step=    5 Error= 4.615E-05
 Mesh structure
 Level  1 has          1 grids (       0,       1,       0,)
...
 Level 15 has        196 grids (      98,      98,      98,)
 Main step=      1 mcons= 0.00E+00 econs= 0.00E+00 epot=-2.27E+07 ekin= 2.76E+03
 Fine step=      1 t= 1.90562E-05 dt= 1.462E-05 a= 1.000E+00 mem=14.7% 
 Time elapsed since last coarse step:    0.01 s       11.88 mus/pt       11.88 mus/pt (av)
 Used memory:     30.4 MB
 Total running time:   5.99999987E-02 s
   ==> Level=    5 Step=    6 Error= 5.033E-05
   ==> Level=    6 Step=    5 Error= 4.597E-05
   ==> Level=    7 Step=    5 Error= 4.636E-05
   ==> Level=    8 Step=    5 Error= 4.638E-05
   ==> Level=    9 Step=    5 Error= 8.631E-05
   ==> Level=   10 Step=    5 Error= 4.132E-05
   ==> Level=   11 Step=    5 Error= 4.326E-05
   ==> Level=   12 Step=    5 Error= 4.706E-05
   ==> Level=   13 Step=    5 Error= 4.639E-05
   ==> Level=   14 Step=    5 Error= 4.618E-05
   ==> Level=   15 Step=    5 Error= 4.564E-05
 Mesh structure
 Level  1 has          1 grids (       0,       1,       0,)
...
```

rteyssier: It should show something like:
CAUTION: Switching to safe MG mode for level...

In init_poisson, there is:

  allocate(safe_mode(1:nlevelmax))
  safe_mode = .false.

  !--------------------------------
  ! For a restart, read poisson file
  !--------------------------------
  if(nrestart>0)then
     ilun=ncpu+myid+10
     
Isn't it what we want?

tinecolman:
> This is not a proper fix for this problem. I would rather output safe_mode(level) to disk so that it is read back at restart.

I don't see why a level should stay in safe_mode for the entire simulation. The convergence issues I've observed are on initial conditions that have sharp edges or right after a SN. It's not that common, at least in the test suite.

poisson/multigrid_fine_commons.f90
tinecolman: I've never seen this print. It should be there if safe_mode is activated. So why does resetting it fix the test if there is no print...

tinecolman: Ah because verbose=.false.

tinecolman: It can be added to the output somehow, if you think that is better. But again, I don't understand why it should be switched on for the entire simulation when the issue only happens at very specific moments in time (init in this case).

rteyssier: There is no easy answer. If you set safe_mode = .false. at each step, the Poisson solver might get inefficient in cases it should be set to .true. for the entire simulation or for a long period of time. When the grid geometry evolves to a better configuration, it is true we could try again and check if safe_mode could be set back to .false. but this is very hard to predict in practice. There is a long discussion in Guillet & Teyssier about this. This is why we decided not to put safe_mode in the restart so that the code tries with safe_mode=.false. every now and then.

tinecolman: I see. Maybe instead of silently turning it back off on restart, we can add a parameter that is going to reset safe_mode every X amount of time steps?

tinecolman:
@rteyssier , I've changed the fix to writing safe_mode to the output and added a way to control after how many solves it should be reset.

What do you want me to do about the restart behavior? As the PR is now, the default behavior will change from "reset on restart" to "restore on restart". An additional namelist flag `mg_reset_safe_mode_on_restart` could be added to keep the old behavior.

Also, maybe you would want the default to change to `mg_safe_mode_reset=10`, as is actually written in Guillet & Teyssier 2011 at the end of section 4?

poisson/poisson_parameters.f90
```diff
@@ -23,6 +23,17 @@ module poisson_parameters
...
+  ! Maximum number of multigrid cycles at a fine level before giving up
+  integer :: mg_maxiter=10
+
+  ! Number of solves for which first-order reconstruction of the multigrid
+  ! boundaries ("safe mode") is kept ON once the solver has switched to it
+  ! for a certain level (counting the solve that triggered the switch)
+  !   1 = switch back after every solve
+  !  10 = good value quoted in Guillet & Teyssier (2011)
+  !   0 = never reset
+  integer :: mg_safe_mode_reset=0
```
rteyssier: Not sure about the best default value here...

tinecolman: Me neither. We could put 10, as in the paper. Something like 100 may be closer to the reset on restart from before.

rteyssier:
Ok let’s put 10.Romain On Sep 24, 2026, at 7:51 AM, Tine Colman ***@***.***> wrote:﻿
@tinecolman commented on this pull request.



In poisson/poisson_parameters.f90:
> @@ -23,6 +23,17 @@ module poisson_parameters
   ! level >=cg_levelmin uses conjugate gradient
   integer :: cg_levelmin=999
 
+  ! Maximum number of multigrid cycles at a fine level before giving up
+  integer :: mg_maxiter=10
+
+  ! Number of solves for which first-order reconstruction of the multigrid
+  ! boundaries ("safe mode") is kept ON once the solver has switched to it
+  ! for a certain level (counting the solve that triggered the switch)
+  !   1 = switch back after every solve
+  !  10 = good value quoted in Guillet & Teyssier (2011)
+  !   0 = never reset
+  integer :: mg_safe_mode_reset=0

Me neither. We could put 10, as in the paper. Something like 100 may be closer to the reset on restart from before.

tinecolman:
Ok, done!
________________________________
Van: Romain Teyssier ***@***.***>
Verzonden: donderdag 24 september 2026 14:52
Aan: ramses-organisation/ramses ***@***.***>
CC: Tine Colman ***@***.***>; Mention ***@***.***>
Onderwerp: Re: [ramses-organisation/ramses] More control over poisson multigrid solver safe_mode (PR #313)

[https://avatars.githubusercontent.com/u/4170009?s=20&v=4]rteyssier left a comment (ramses-organisation/ramses#313)<https://github.com/ramses-organisation/ramses/pull/313#issuecomment-5814473115>
Ok let’s put 10.Romain On Sep 24, 2026, at 7:51 AM, Tine Colman ***@***.***> wrote:﻿
@tinecolman commented on this pull request.



In poisson/poisson_parameters.f90:
> @@ -23,6 +23,17 @@ module poisson_parameters
! level >=cg_levelmin uses conjugate gradient
integer :: cg_levelmin=999

+ ! Maximum number of multigrid cycles at a fine level before giving up
+ integer :: mg_maxiter=10
+
+ ! Number of solves for which first-order reconstruction of the multigrid
+ ! boundaries ("safe mode") is kept ON once the solver has switched to it
+ ! for a certain level (counting the solve that triggered the switch)
+ ! 1 = switch back after every solve
+ ! 10 = good value quoted in Guillet & Teyssier (2011)
+ ! 0 = never reset
+ integer :: mg_safe_mode_reset=0

Me neither. We could put 10, as in the paper. Something like 100 may be closer to the reset on restart from before.

rteyssier (approved):

Issue #121, opened 2025-06-11
Simulation with Monte Carlo tracers fail at restart


nbrucy:
## How to reproduce

Run the MC tracer test  with the restart option activated.
In `tests`,
```
./run_test_suite.sh -rv -p 2 -t tracer
```

## Expected behaviour

The test passes

## Actual result

The test fail. Actually, ramses doesn't even restart because of a "particle outside boundary" error.

```
 Problem in check_tree at level            5
 A particle has moved outside allowed boundaries
   7.8125000000000000E-003  0.25781250000000000     
  0.43750000000000000       0.21875000000000000    
```

nbrucy:
The bug is not reproduced anymore, so closing.

tinecolman:
Turns out the test passed because the tolerances where set very high and there was no check on the time. So it passed with restart even though it didn't reach the end of the simulation...

tinecolman:
The cause of this is that on restart `initialize_skip_loc` is not called before the `scale` variable that is defined in the `tracer_utils` module is used.

PR #314, opened 2026-08-21
Fix crash on restarts for MC tracers

tinecolman:
The cause of this is that on restart `initialize_skip_loc` is not called before the `scale` variable that is defined in the `tracer_utils` module is used.

Fixed by calling it at the start of `init_part`.

Also fixed a wrong partp due to iskip not being recalculated in the loop. Though this does not seem to have an impact since it is recalculated later on.


tinecolman:
This fixes issue #121

nbrucy (approved):
It indeed solves the issue of MC tracers crashing at restart.

PR #317, opened 2026-08-31
Option to restart with sinks from an output that did not have sinks

florian-ded:
The current version of RAMSES does not allow to restart with the option sink = .true. from an output of a simulation that did not implement sinks (so with no sink file in the output). More precisely, the init_sink subroutine of init_sink.f90 tries to open a sink file and read 2 lines that do not exist. I added a few lines that create the file if it does not exist and write the necessary 2 lines. The MPI implementation is taken into account and in particular CPU are synchronized to make sure that they can all open and read the file after the lines were written. The automatic test suite on github ran successfully.


rteyssier (changes requested):
What are these large binary files?
You should not put this in a git repo.

nbrucy:
Thanks for your PR Florian.
The large binary files are the self-generated test ic files and has been added by mistake. @florian-ded you can remove them with `git rm` and update your PR.
We can then run the the autofix for the pre-commit bot.

florian-ded:
Hello Romain and Noé,
I removed the binary files, it should be ok now

pm/init_sink.f90
```diff
@@ -138,6 +138,22 @@ subroutine init_sink
 #endif
 
      nsink=0
+     ! Create an empty sink file if the previous snapshot had no sinks.
+#ifndef WITHOUTMPI
+     if ((IOGROUPSIZE == 0 .and. myid == 1) .or. (IOGROUPSIZE > 0 .and. mod(myid-1,IOGROUPSIZE) == 0)) then
+#endif
+        inquire(file=fileloc, exist=ic_sink)
```
nbrucy: ic_sink has another meaning in the routine (whether there is an initialisation file for sink).
For code readability, maybe declare a new boolean with a distinct name eg `exists_sink_restart`.

nbrucy (changes requested):

nbrucy (approved):

pm/init_sink.f90
```diff
@@ -12,7 +12,7 @@ subroutine init_sink
 #endif
   real(dp)::scale_nH,scale_T2,scale_l,scale_d,scale_t,scale_v
   integer::isink, nsinkold
-  logical::eof,ic_sink=.false.
+  logical::eof,ic_sink, exists_sink_restart=.false.
```
rteyssier, line 15: Remove unused ic_sink logical.

rteyssier (changes requested):

pm/init_sink.f90
nbrucy: It is used later down in the file (for when you read a file to initialize anew sinks).

rteyssier: Ah OK my bad!

rteyssier (approved):

pm/init_sink.f90
mtrebitsch: Shouldn't we keep the declaration `ic_sink=.false.` as well then?
It is set by https://github.com/ramses-organisation/ramses/blob/bcb8de501edf22f64781bea746e4198f731ab930/pm/init_sink.f90#L211
so this should always be set, but maybe we should stay on the safe/explicit side?

nbrucy: Actually, the initialisation is only done the first time the function is called (yeah fortran is weird). So the initialisation to .false. in the declaration part is quite useless in this context, the only way to be sure it's initialized is to put `ic_sink=.false.` in the body of the routine.

PR #324, opened 2026-09-16
Change the way the particle linked lists are updated

tinecolman:
**Issue**
The routines `kill_tree_fine` and `make_tree_fine` modify the particle linked lists *while* walking the list. This presents an obstacle for adding openmp.

**Solution**
First record which particles need to move during a first pass through the tree, then apply the changes during a second pass.

For this, two new arrays are added:
* `newgridp(1:npartmax)` which records to grid to which each particle needs to move
* `nmovep(1:ngridmax)` which records how many moves to do for each grid (optimisation purposes)
The routines `kill_tree` and `check_tree` update these arrays instead of altering the linked list.
A new routine, `apply_tree_moves`, will apply the moves recorded in these arrays to the linked list. A check on `nmovep` allows to skip part of the tree walk when no changes are needed.

**OpenMP**
With this implementation, the most expensive parts can be parallelised with openmp, while only `apply_tree_moves` needs to be performed in serial. This way, the particle order is preserved, with is important for reproducing MC tracers.

**Side effect**
The calls to `remove_list` and `add_list` by the routine `kill_tree` and `check_tree` were the only ones that actually made use of the `ok` flag. All other call just pass a hard-coded `true` list. Since now the `ok` flag information has already been incorporated in `newgridp`, the ok-check is no longer needed. It has been removed.
Similarly, it turned out that the ok flag in `add_free_cond` was never actually filled with something useful. This routine has been removed, in favor of using the regular add_free.

**Performance**
I verified this doesn't impact performance by comparing the times for the benchmarks cosmo, cosmo-amr and galaxy, on 1 to 4 nodes of MeluXina. The execution times are indistinguishable between old and new implementation.


amr/end.f90
```diff
@@ -122,6 +122,8 @@ subroutine deallocate_pm
      if(allocated(mp)) deallocate(mp)
      if(allocated(vp)) deallocate(vp)
      if(allocated(xp)) deallocate(xp)
+     if(allocated(newgridp)) deallocate(newgridp)
+     if(allocated(nmovep)) deallocate(nmovep)
```
rteyssier: This adds non negligible memory usage.

tinecolman: Yes. In principle it can be done without the extra grid array nmovep. I can run a test to estimate how much that would affect performance.
Another option is to put this behind an openmp wall, but that will make the code complicated.

PR #328, opened 2026-09-21
Fix bugs related to level-dependent speed of light

tinecolman:
A few bugs were introduced when the level-specific speed-of-light was added.

**rt refinement flagging:**
* The vector-sweep variable `c_factor` is initialized only once before the loop and then updated when a neighbor doesn't exist. However the updated value is never reset back to the default and wrongly persists for the next vector sweep.
* The type of `c_factor` is set to integer while `rt_c` is a double, causing it to be truncated.

**Outputting:**
* When ilevel is lower than levelmin,` rt_c(ilevel)=0`. By luck, this is mostly harmless. However, it causes wrong outputting and restarting for cells at levels < levelmin. For those cells, the photon density is always outputted as zero and read back as `0/0=NaN`.

This was not caught by tests because 
1) there is no test which uses the rt refinement criterion,
2) in post-processing, we usually don't read cells at level<levelmin.

Adding a test which uses amr with level-dependent speed of light revealed more bugs.

**Interpolation between levels**
No correction for rt_c(level) is applied during the restriction. This means that cells at the *same* level can carry *different* rt_c units, depending on whether they have children. Only leaf cells are at their own rt_c(level). Non-leaf cells are in some mix of [rt_c(level+1),rt_c(level+2),...] depending on its descendants. This makes the whole thing difficult to keep track off. The correction wasn't applied consistently. This caused issues at the boundaries which distorted the stromgren sphere.
To fix this, the correction is now applied directly in rt_upl, so that all cells are at the correct rt_c(level).

**Stale boundaries**
The rt refinement criterion needs the updated information in the boundaries. However those boundaries where not updated after the call to rt_upload_fine, unlike their hydro counterparts and so the stale information was used. This caused the solution for the test to vary with ncpu.
To fix this, an rt boundary communication and update has been added in amr_step after the corresponding hydro block.

**Other bugs unrelated to the level-dependent speed of light**
Passing the thing to Claude Code for review revealed 2 more pre-existing bugs.
* The units for the flux in a square rt sources had a factor c too many, causing the value to be of the order of c instead of between 0 and 1. This is later clipped to 1, so anything that used `rt_u_source` between 0 and 1 was actually using 1. Remark that these parameters are poorly documented.
* Missing parentheses in a reduction factor in coolfine1, when using rt_vc.


jokirosdahl (approved):
Looks good to me!
Thanks Tine for spotting these issues.

tinecolman:
While modifying the stromgren2d-He test to use the level-dependent speed of light, I discovered another more serious issue. The result I got was this:
Screenshot from 2026-09-21 19-00-38

Following this rabbit hole, lead to the finding that cells in a certain level can have inconsistent rt_c units depending on whether they are leaf cells or not. This caused the rt_c correction to sometimes be applied twice.
With the fix, all cells should now consistently be at their rt_c units since the value is corrected when doing the restriction.

This restores the shock front to its circular shape.

tinecolman:
The new test surfaced another issue where stale boundary values are used in the refinement check, and so the test solution varied with ncpu.

jokirosdahl:
Wow! These changes make sense for me. I wonder if it would make the whole thing much easier if we stored flux, i.e. c*N, instead of the photon density N. Then no rescaling would be needed.

It is probably a good idea to use refinement in both stromgren2d tests and refine on both photon flux (err_grad_cn) and ionization fraction (e.g. rt_err_grad_xHI). I remember refining on those tended to break the tests when running with the variable speed of light.

tinecolman:
> It is probably a good idea to use refinement in both stromgren2d tests and refine on both photon flux (err_grad_cn) and ionization fraction (e.g. rt_err_grad_xHI). I remember refining on those tended to break the tests when running with the variable speed of light.

Indeed, the broken stromgren sphere I got used `rt_err_grad_cn=0.4`. I'll try the other one.

tinecolman:
I've investigated a bit more. On dev, the non-spherical front bug is hidden behind the first bug in the refinement flagging. Because of the wrong type of c_factor, the left and right side of the gradient calculation are always 0 while the center is non-zero. This causes the whole zone which contains photons to be refined. The refinement boundary is in a zone where the photon density is 0 and so the rt_c scaling problem doesn't show up because scaling 0 is just 0. When correcting c_factor to be a double, there is less refinement and the refinement boundary shifts to a zone where the rt_c scaling does matter.

So indeed, whether the bug showed or not depended on the refinement strategy.

jokirosdahl:
> So indeed, whether the bug showed or not depended on the refinement strategy.

But is it behaving now, with your fixes, when refining on the photon flux?

tinecolman:
> But is it behaving now, with your fixes, when refining on the photon flux?

[test_results.pdf](https://github.com/user-attachments/files/32517903/test_results.pdf)
These are the results with 
```
levelmin=7
levelmax=8
rt_c_fraction=0.001,0.0005
```
and `rt_err_grad_xHI=0.2`  for stromgren2d and `rt_err_grad_cn=0.4` for stromgren2d-He.

tinecolman:
Two more bugs:
* the units for the flux in a square rt sources had a factor c too many, causing the value to be of the order of c instead of between 0 and 1. This is later clipped to 1, so anything that used `rt_u_source` between 0 and 1 was actually using 1. Remark that these parameters are poorly documented.
* Missing parentheses in a reduction factor in coolfine1, when using rt_vc.

These seem minor to me but they are not covered by any tests, so I can't be sure what the impact would be.

tinecolman:
I hope I'm done now!
This all started because `c_factor` being initialized outside the vector sweep was a problem for openmp...

jokirosdahl:
> I hope I'm done now! This all started because `c_factor` being initialized outside the vector sweep was a problem for openmp...

Well done! Thanks a lot!

rteyssier (approved):

